PR #3, opened 2015-07-01
Made status matching stricter to reject false matches

webmozart:
The bot now only matches a status change as status change if it's written on a single line. The reason is that if a comment quotes a label or explains how to change a status (see test cases), then the bot should not actually change the status.


weaverryan:
Really good thinking - test cases made it very clear - less WTF :)

PR #22, opened 2015-12-16
Fix standards and update dependencies

Ener-Getick:
This PR:
- [x] updates the dependencies
- [x] fixes the PHP coding standards (thanks to ``PHP-CS-FIXER``)

PR #23, opened 2015-12-16
Allow to manage several repositories at the same time

Ener-Getick:
This PR includes #22 and adds support for multiple repositories.


weaverryan:
@Ener-Getick Can you rebase now that #22 is merged? It'll just make reviewing easier.

Thanks!

Ener-Getick:
Done @weaverryan

src/AppBundle/Controller/DefaultController.php
```diff
@@ -1,19 +0,0 @@
-<?php
-
-namespace AppBundle\Controller;
-
-use Symfony\Bundle\FrameworkBundle\Controller\Controller;
-use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
-
-class DefaultController extends Controller
```
Ener-Getick: I removed this controller as it is not really useful and would be complicated to manage with many repositories.

src/Repository/Provider/InMemoryRepositoryProvider.php
```diff
@@ -0,0 +1,52 @@
...
+
+    /**
+     * @param string $owner
+     * @param string $name
+     *
+     * @return Repository|null
+     */
+    public function getRepository($owner, $name)
+    {
+        if (isset($this->repositories[strtolower($owner.'/'.$name)])) {
```
sstok: You can cache the strtolower() result in variable to safe a method call.

Ener-Getick: Fixed

Ener-Getick:
No more reviews?

app/config/config.yml
```diff
@@ -7,6 +7,8 @@ imports:
 # http://symfony.com/doc/current/best_practices/configuration.html#application-related-configuration
 parameters:
     locale: en
+    repositories:
+        - { owner: %repository_username%, name: %repository_name% }
```
WouterJ: Both values should be single-quoted strings, as `%` is a reserved Yaml character.

src/AppBundle/Controller/WebhookController.php
```diff
@@ -18,8 +18,18 @@ public function githubAction(Request $request)
     {
         $data = json_decode($request->getContent(), true);
         if ($data === null) {
-            throw new \Exception('Invalid JSON body!');
+            return new JsonResponse('Invalid JSON body!', 400);
```
WouterJ: This should still be an exception imo.

src/AppBundle/Controller/WebhookController.php
```diff
@@ -18,8 +18,18 @@ public function githubAction(Request $request)
...
         }
+        if (!isset($data['repository']['full_name'])) {
```
WouterJ: missing empty line before this if

src/AppBundle/Controller/WebhookController.php
```diff
@@ -18,8 +18,18 @@ public function githubAction(Request $request)
...
+            return new JsonResponse('Repository name not defined!', 400);
```
WouterJ: should be exception as well imo

src/AppBundle/Controller/WebhookController.php
```diff
@@ -18,8 +18,18 @@ public function githubAction(Request $request)
...
+        }
+
+        list($owner, $name) = explode('/', $data['repository']['full_name']);
```
WouterJ: what about using `vendor`, `org`, etc. instead of owner?

src/AppBundle/Controller/WebhookController.php
```diff
@@ -18,8 +18,18 @@ public function githubAction(Request $request)
...
+        $repository = $this->get('repository_provider')->getRepository($owner, $name);
+        if ($repository === null) {
```
WouterJ: Should be `null === $repository`

src/AppBundle/Controller/WebhookController.php
```diff
@@ -18,8 +18,18 @@ public function githubAction(Request $request)
...
+            throw $this->createNotFoundException();
```
WouterJ: let's add a usefull message?

src/Repository/Repository.php
```diff
@@ -0,0 +1,76 @@
...
+{
+    /**
+     * @var string
+     */
+    private $owner;
+    /**
+     * @var string
+     */
+    private $name;
+    /**
```
WouterJ: there should be an empty line before the PHPdoc

src/AppBundle/Issues/GitHub/GitHubStatusApi.php
```diff
@@ -90,10 +85,15 @@ public function getIssueStatus($issueNumber)
 
     public function getNeedsReviewUrl()
     {
+        $repository = $this->repositoryStack->getCurrentRepository();
+        if ($repository === null) {
```
WouterJ: `null === $repository`

src/AppBundle/Issues/GitHub/CachedLabelsApi.php
```diff
@@ -93,4 +91,14 @@ public function removeIssueLabel($issueNumber, $label)
             unset($this->labelCache[$issueNumber][$label]);
         }
     }
+
+    private function getRepository()
+    {
+        $repository = $this->repositoryStack->getCurrentRepository();
+        if ($repository === null) {
```
WouterJ: `null === $repository`

WouterJ: as this code is duplicated a lot of times, why not add it to the method?

app/config/config.yml
Ener-Getick: Fixed thanks :-)

src/AppBundle/Controller/WebhookController.php
Ener-Getick: Github uses ``owner`` (see [here](https://github.com/carsonbot/carsonbot/blob/master/src/AppBundle/Tests/webhook_examples/pull_request.opened.json#L25)) but I agree that ``vendor`` seems to be better.

src/AppBundle/Controller/WebhookController.php
Ener-Getick: You're right, would be better to create an exception controller later.

Ener-Getick:
Done @WouterJ

WouterJ:
@weaverryan please have a look at this. This feature is required if we want to use Carson for the symfony docs.

src/AppBundle/Controller/WebhookController.php
```diff
@@ -21,6 +21,17 @@ public function githubAction(Request $request)
...
         }
 
+        if (!isset($data['repository']['full_name'])) {
+            throw new \Exception('Repository name not defined!');
+        }
+
+        list($vendor, $name) = explode('/', $data['repository']['full_name']);
+        $repository = $this->get('repository_provider')->getRepository($vendor, $name);
+        if (null === $repository) {
+            throw $this->createNotFoundException(sprintf('Repository %s/%s not found.', $vendor, $name));
```
WouterJ: "not found" sounds like it doesn't exists (on GitHub). What about saying "Repository "%s/%s" is not enabled." or "Repository "%s/%s" is not supported"?

src/Repository/Provider/InMemoryRepositoryProvider.php
```diff
@@ -0,0 +1,52 @@
...
+            return $this->repositories[$fullName];
+        }
+    }
+
+    /**
+     * Adds a repository.
+     *
+     * @param array $options
+     */
+    public function addRepository(array $options)
```
WouterJ: any reason to use a very generic `$options` above a more strict and self explaining `$vendor, $name, array $options = []` arguments?

src/Repository/Provider/InMemoryRepositoryProvider.php
```diff
@@ -0,0 +1,52 @@
...
+        foreach ($repositories as $repository) {
+            $this->addRepository($repository);
+        }
+    }
+
+    /**
+     * @param string $vendor
+     * @param string $name
+     *
+     * @return Repository|null
```
WouterJ: I think it's better to throw an exception if the repository is not disabled. I can't imagine much use cases to return null.

src/Repository/Provider/InMemoryRepositoryProvider.php
```diff
@@ -0,0 +1,52 @@
...
+     */
+    public function getRepository($vendor, $name)
+    {
+        if (isset($this->repositories[$fullName = strtolower($vendor.'/'.$name)])) {
+            return $this->repositories[$fullName];
+        }
+    }
+
+    /**
+     * Adds a repository.
```
WouterJ: not so usefull PHPdoc description. Imo, this one can be removed.

src/Repository/Provider/InMemoryRepositoryProvider.php
```diff
@@ -0,0 +1,52 @@
+<?php
+
+namespace Repository\Provider;
```
WouterJ: why doesn't this live in `AppBundle\Repository\Provider`?

src/Repository/Repository.php
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * @var string
+     */
+    private $name;
+
+    /**
+     * @var string|null
+     */
+    private $secret;
```
WouterJ: what's the exact use-case of `$secret` here? I can't see it being used.

src/Repository/RepositoryStack.php
```diff
@@ -0,0 +1,46 @@
...
+    {
+        if (!$this->repositories) {
+            return;
+        }
+
+        return array_pop($this->repositories);
+    }
+
+    /**
+     * @return Repository|null
```
WouterJ: this will never return `null`, an exception would be thrown otherwise.

src/Repository/RepositoryStack.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+namespace Repository;
+
+/**
+ * @author Ener-Getick <egetick@gmail.com>
+ */
+class RepositoryStack
```
weaverryan: We don't really need a "stack", do we? We will only every be dealing with *one* repository at a time. What about `RepositoryContext`, with a `setCurrentRepository` and `getCurrentRepository()`? If there is no current repository and someone calls `getCurrentRepository`, we should throw an exception (as that's probably a developer error).

weaverryan:
Sorry for the delay reviewing this @Ener-Getick - I wasn't interested in having this for multiple repositories before now. We would like to enable this also for the `symfony-docs` repository. I just added one main comment, but this looks like a very sane step to make that easily possible.

We will also add some symfony-docs-only functionality. The new `Repository` class will make it very easy to configure different feature flags that each has in the future.

Ener-Getick:
No problem @weaverryan, i won't be able to fix your comments before tonight or maybe tomorrow so in case you want to be quick, i gave you the rights on my repository.

src/Repository/RepositoryStack.php
Ener-Getick: I was thinking about interpedent repositories (for example, carsonbot could move the pull requests in the main repository). A stack would allow to fetch data from the sub repo, then change the current repository, then call the github api and then pop the current repository to do other threatments. We can also add this later it doesn't matter too much.

src/Repository/Repository.php
Ener-Getick: None for now.
I wanted to securize the api (anyone who knows its url can alter sf pull requests) but i decided to do it later.
We can remove this for now.

src/Repository/Provider/InMemoryRepositoryProvider.php
Ener-Getick: I think this is somehow generic so could be in the root namespace. But you're right, that's not coherent with the rest of the code.

src/Repository/Provider/InMemoryRepositoryProvider.php
Ener-Getick: It's to be coherent with the constrictor. In fact we could even use the Repository class here.

src/Repository/RepositoryStack.php
HeahDude: Yes I saw that comment, but I don't understand
>interpedent repositories (for example, carsonbot could move the pull requests in the main repository). A stack would allow to fetch data from the sub repo

What do you mean by that?

Because it seems to me that this repo is an api for post request and each of them is bound to one repo.

If you're thinking of read-only sub like https://github.com/silexphp/Silex-Api, I guess only current users can tell if it might be a use case for it in the future.

In any case this could be done easily after #31 as well.

From this PR how would you handle the labels "hasPR" and "in progress" used for the docs and not in the core for example, I think it requires repo specific actions, right?

Ener-Getick: Yeah this is what i mean, sorry if it wasn't clear.
This PR only proposes a solution for managing multiple repositories not for managing them differently.
But as i said in your PR, the EventDispatcher is imo what we need and we could create a features or listeners or whatever property in the Repository class to enable or disable a listeners based on the current repository.

Of course this PR is not as complete as yours (the event management could be used here too) and i won't have the time before wednesday to do big changes so i let @WouterJ and @weaverryan decide which approach they prefer.

Thanks for your review and your PR ☺

Ener-Getick:
I fixed your comments @WouterJ.
@weaverryan with my answer, do you still think we should replace the ``RequestStack`` ?

Ener-Getick:
Closed in favor of https://github.com/carsonbot/carsonbot/pull/33

PR #29, opened 2016-03-07
Autolabel new Pull Requests

javiereguiluz:
Two important things:

1) I haven't tested this in any way and I don't know how to test it in a real repository.

2) ~~What happens if you try to label an issue with a non-existent label? I guess it will be dropped silently. Otherwise, we'll have a problem if a user creates a PR with a title like this: "[Whatever] Blah blah"~~ FIXED in the latest version of this PR.

### DIFF
--- a/src/AppBundle/Issues/IssueListener.php
+++ b/src/AppBundle/Issues/IssueListener.php
@@ -129,10 +129,10 @@ class IssueListener
         $labels = array();
 
         // e.g. "[PropertyAccess] [RFC] [WIP] Allow custom methods on property accesses"
-        if (preg_match_all('/\[(?P<tags>.+)\]/U', $prTitle, $matches)) {
-            foreach ($matches['tags'] as $tag) {
-                if (in_array($tag, $this->getValidTags())) {
-                    $labels[] = $tag;
+        if (preg_match_all('/\[(?P<labels>.+)\]/U', $prTitle, $matches)) {
+            foreach ($matches['labels'] as $label) {
+                if (in_array($label, $this->getValidLabels())) {
+                    $labels[] = $label;
                 }
             }
         }
@@ -141,9 +141,9 @@ class IssueListener
     }
 
     /**
-     * TODO: get valid tags from the repository via GitHub API
+     * TODO: get valid labels from the repository via GitHub API
      */
-    private function getValidTags()
+    private function getValidLabels()
     {
         return array(
             'Asset', 'BC Break', 'BrowserKit', 'Bug', 'Cache', 'ClassLoader',
--- a/src/AppBundle/Issues/IssueListener.php
+++ b/src/AppBundle/Issues/IssueListener.php
@@ -71,9 +71,9 @@ class IssueListener
         $newStatus = Status::NEEDS_REVIEW;
         $prLabels[] = $newStatus;
 
-        // the PR title usually indicates the affected component
-        if (preg_match('/^\[(?P<component>.*)\] .*/$', $prTitle, $matches)) {
-            $prLabels[] = $matches['component'];
+        // the PR title usually contains one or more labels
+        foreach ($this->extractLabels($prTitle) as $label) {
+            $prLabels[] = $label;
         }
 
         // the PR body usually indicates if this is a Bug, Feature, BC Break or Deprecation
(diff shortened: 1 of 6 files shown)


src/AppBundle/Issues/IssueListener.php
```diff
@@ -58,14 +58,39 @@ public function handleCommentAddedEvent($issueNumber, $comment)
      * Adds a "Needs Review" label to new PRs.
      *
      * @param int $prNumber The number of the PR
+     * @param int $prTitle  The title of the PR
+     * @param int $prBody   The full text description of the PR
      *
      * @return string The new status
      */
-    public function handlePullRequestCreatedEvent($prNumber)
+    public function handlePullRequestCreatedEvent($prNumber, $prTitle, $prBody)
```
webmozart: If we listen to this event, we should also listen for changes to the title/body after the PR was created.

javiereguiluz: Probably ... but we can leave that enhancement for the future.

src/AppBundle/Issues/IssueListener.php
```diff
@@ -58,14 +58,39 @@ public function handleCommentAddedEvent($issueNumber, $comment)
...
     {
+        $prLabels = array();
+
+        // new PRs always require review
         $newStatus = Status::NEEDS_REVIEW;
+        $prLabels[] = $newStatus;
+
+        // the PR title usually indicates the affected component
```
WouterJ: Assuming the first bracket will be the component name is far from true. Usually, `[WIP]`, `[RFC]`, `[POC]` are put before the change. Even more, some people tend to include the version number in the PR title, also within brackets. This is placed before the component brackets as well.

Imo, we should get a list of available labels for a repository and check all brackets as long as the value inside the brackets do not match a label

src/AppBundle/Issues/IssueListener.php
```diff
@@ -58,14 +58,39 @@ public function handleCommentAddedEvent($issueNumber, $comment)
...
+        if (preg_match('/^\[(?P<component>.*)\] .*/$', $prTitle, $matches)) {
+            $prLabels[] = $matches['component'];
+        }
+
+        // the PR body usually indicates if this is a Bug, Feature, BC Break or Deprecation
+        if (preg_match('/^\|\s*Bug fix?\s*\|\s*yes\s*$/', $prBody, $matches)) {
```
WouterJ: `\s*` should be replaced by `\s+` around the middle `|`

src/AppBundle/Issues/IssueListener.php
```diff
@@ -58,14 +58,39 @@ public function handleCommentAddedEvent($issueNumber, $comment)
...
+    public function handlePullRequestCreatedEvent($prNumber, $prTitle, $prBody)
     {
+        $prLabels = array();
+
+        // new PRs always require review
         $newStatus = Status::NEEDS_REVIEW;
+        $prLabels[] = $newStatus;
+
+        // the PR title usually indicates the affected component
+        if (preg_match('/^\[(?P<component>.*)\] .*/$', $prTitle, $matches)) {
```
WouterJ: the last `.*` seems not needed and only slow down the operation.

src/AppBundle/Issues/IssueListener.php
```diff
@@ -58,14 +58,39 @@ public function handleCommentAddedEvent($issueNumber, $comment)
...
+            $prLabels[] = $matches['component'];
+        }
+
+        // the PR body usually indicates if this is a Bug, Feature, BC Break or Deprecation
+        if (preg_match('/^\|\s*Bug fix?\s*\|\s*yes\s*$/', $prBody, $matches)) {
+            $prLabels[] = 'Bug';
+        }
+        if (preg_match('/^\|\s*New feature?\s*\|\s*yes\s*$/', $prBody, $matches)) {
+            $prLabels[] = 'Feature';
+        }
```
WouterJ: do we want to allow both `Bug` and `Feature` labels on one PR?

javiereguiluz: Yes. Here it is an example: https://github.com/symfony/symfony/pull/17458

src/AppBundle/Issues/IssueListener.php
javiereguiluz: `\s*` gives us more flexible "for free".

src/AppBundle/Issues/IssueListener.php
webmozart: Sorry if I'm repeating myself :) Why not check the paths of the modified files? Isn't that much more reliable?

src/AppBundle/Issues/IssueListener.php
WouterJ: `\s*` is incorrect, there is no reason to allow incorrect flexibility

javiereguiluz: This is my reasoning. If some user writes this by mistake:

```
| Bug fix? |yes
```

`\s*` will work but `\s+` won't. I can be wrong of course.

src/AppBundle/Issues/IssueListener.php
javiereguiluz: It's more reliable, but much more complicated. I'm trying to follow the MVP philosophy:

1) Today we have "nothing".
2) This PR adds "something" (basic autolabel based on title+description).
3) In the future we'll have "everything" (smart autolabel based on modified files).

In my experience, if you try to have "everything" before "something", you'll end up with "nothing".

webmozart: Fair enough :) I'm questioning whether it really is that complicated. As far as I saw, it's an API call (max) + checking modified paths, which is only marginally more complicated than the code we have here.

src/AppBundle/Issues/IssueListener.php
javiereguiluz: You are right. That regular expression was awful. I've updated everything. Thanks.

src/AppBundle/Issues/IssueListener.php
javiereguiluz: > Imo, we should get a list of available labels for a repository and check all brackets as long as the value inside the brackets do not match a label

You are right ... but I prefer to leva that for the future. For now I've hardcoded the valid labels, which sounds awful ... but it's not that bad because they rarely change.

javiereguiluz:
If this feature is finally approved, should we ...

* ... merge it as soon as possible to confirm is ready for the Hack Day of this Saturday?
* ... or merge it *after* this Saturday to not interfere with the Hack Day?

Thanks!

javiereguiluz:
Friendly ping to not forget about this PR. What could we do to move it forward? Thanks!

sstok:
@javiereguiluz Tests are missing, overall it looks pretty good :+1:

javiereguiluz:
@sstok thanks for the review. I've fixed the failing test.

weaverryan:
See #34 instead. @javiereguiluz we made such big changes to the repo, it wasn't fair to have you try to figure out the conflict :)

Issue #30, opened 2016-03-07
Use the EventDispatcher to split the logic in the WebhookController


Ener-Getick:
The [``WebhookController``](https://github.com/carsonbot/carsonbot/blob/master/src/AppBundle/Controller/WebhookController.php) becomes more and more complex so I think we should use the ``EventDispatcher`` to simplify it and create a new ``github.hook`` event:
```php
class CommentSubmissionListener implements EventSubscriberInterface {
    public function onGithubHook(HookEvent $event) {
        if ('issue_comment' !== $event->getEvent()) {
            return;
        }
        // ...

        $event->addReturn('Comment managed');
    }

    public static function getSubscribedEvents() {
        return array('github.hook' => 'onGithubHook');
    }
}
```

What do you think ?

PR #31, opened 2016-05-08
Refactored event handling

HeahDude:
An alternative approach of #23.

This is a discussion opening :) I think we should build dynamically the listener depending on the repository. IMHO, this way allows to ease specific configuration for one or many repositories.

A global events class list all GitHub events so we can hook any of them in the future easily.

It should not impact #29 too much while rebasing as far as I can see, only the change in `WebhookController.php` is not needed since data is accessible directly in the listener method through the event.

What do you think? Thanks!


Ener-Getick:
:+1: for using the dispatcher but :-1: for having different listeners (i prefer choosing the github events in the config instead of hardcoding it, which goes in the way of #23).

Ener-Getick:
BTW this also fixes https://github.com/carsonbot/carsonbot/issues/30.

HeahDude:
Thanks @Ener-Getick for the review :) I realise I didn't really argue in the description, so let's do it.

Aside:
First, it's just a coincidence, I found yesterday https://github.com/sensiolabs/dx.symfony.com and as it is for me an example of usage of https://github.com/KnpLabs/php-github-api and I may need it in a project I'm working on, I decided to first try to make a lightweight "GitHub Issues Manager", and while surfing on GitHub I found this repo and "watched" it to come back to it later. But this morning with the notification from @weaverryan on your PR, I took a look :) I was not very happy with your implementation and totally agree with his comment.
So I tried to work on another approach. Then I've only quickly checked other opened PRs (#29) but not the issues, so I was not aware of #30. Glad that fixes it ;)
It was just a morning exercise for me, it's not intended to be merged, only to discuss another POV :)
That being said.
___________

This PR does:

 - replace `app.status_api` by `app.status_api.default`, preventing the deletion of the `DefaultController` (in contrary of #23).

 - allow a simple parameters config with:

   - `default_repository` used for `app.status_api.default`,

   - `github_listeners`, an array of `$repo => $listenerClass` (e.g `'symfony/symfony' => '\AppBundle\Listener\SymfonyListener``,

 - for now keep the "global logic" in `IssueListener` made abstract so any listener using global event logic can easily reuse it and be tested (WIP on tests).

_________

1. I think it's good to have one listener with specific events subscribing and handling for each repo, because they may not share the same needs (e.g events, data, config with specific titles and labels) (we could improve the collect and configuration compared to what's proposed in that PR).

2. Each request deals with one repo at the time so dynamically generate the right listener using a factory like here is just one possible way. I can't see the use case for a stack, please enlighten me!

3. I believe it can ease maintainability and extendability in the future (e.g #29 and @webmozart's comments), if we intend to handle more repos (e.g SE, demo). Also by fixing #30.

4. A thought: lot of AppBundle files (maybe all but listeners) could go in a specific carson/github-bundle for reusability, could be done in another PR.

Right now, I'm playing with the tests, so I may push them when I'm done with it eventually.

Thanks for sharing your thoughts!

Ener-Getick:
I answered to @weaverryan in #23 about the stack https://github.com/carsonbot/carsonbot/pull/23#discussion_r62427864. We don't need it yet but it could be the case in the future.

Ener-Getick:
And when i designed #23, i thought about creating a library (with a Repository class).
I prefer when my code is not too much coupled with a framework.

Anyway i completely agree with you for using the dispatcher but i don't like much the listener registration at runtime and having a code with limitations (see my message above).

HeahDude:
Ok, thanks for your comments.

I just don't understand what you mean by "hardcoding". Could you please precise your thought?

Ener-Getick:
I meant creating a listener/class for each repository (i prefer using the configuration here)

HeahDude:
Arf, I assumed that configuring by mapping a repo for a **subscriber** class in a parameter would be enough here.

Also, what I missed to say is that `IssueListener` is made abstract **AND** implements `EventSubscriberInterface`.
So each subscriber (not listener) implementation extends global events methods, but needs to define the registered event. You can then easily override (and call parent) plus add specific event methods.

What kind of configuration would you expect in addition here ? Thanks for your answers so far :)

HeahDude:
Ok, I've done some minor fixes and pushed 3 other commits:

1. 5c1ef8e Refactored event handling (fixed),

2. 3f8cca4 Updated composer dependencies and added `phpunit` in `require-dev`,

3. 0ca380a Refactored tests,

4. e6c22cd Added an exception listener to help debugging.

5. If reviews are favorable I'll add more tests for exception cases.

Thanks!

Ener-Getick:
I was thinking about something like:
```yml
repositories:
    symfony/symfony: [ review-statuses, component-labels ]
    symfony/symfony-doc: [ doc-review-statuses ]
```

The listeners would then be automatically configured based on this.

HeahDude:
What would `review-statuses` be used for?

Ener-Getick:
It's just pseudo code, i'm not really satisfied by it but i would use this kind of structure.
``review-statuses`` targets the listener managing the labels (i shouldn't have used ``statuses``) of the main symfony repository.

HeahDude:
Ok, I've refactored after good points of @Ener-Getick about using many listeners for one repo and security.

There is now on main service call from the controller `GitHubRequestHandler::handle($request)` to get the response data.

The handle takes the labels api, the dispatcher and the repositories config as constructor arguments.
It checks request conformity, instantiates the right listeners, and dispatches the event.

I found it very clear now, with specific exceptions to help debugging, sending the right json data and response codes on errors.

Tests are only missing the handler implementation and exceptions behavior, but this globally tested by the controller's tests, which I can't execute in totality because of wrong credentials on the api.
Will further work on it, if the implementation is validated.

Previous commits are outdated and intended to be squashed, also I've reverted the composer update (It should really be done in another PR imho).

I suggest reviewing the global diff though. Thanks!

HeahDude:
I've pushed d4b1030 to try a way to implement a symfony-docs listener. It may help to define proper  way to handle many repositories. Looking at this proposal one might better know what he wants and does not want to do/use.

I've also attached the repository name and its maintainers alias to the event to take advantage of "private" hooks.

Thanks.

Ener-Getick:
@HeahDude your PR becomes really big, imo it would be better to split it to simplify reviews ;-)

HeahDude:
closing in favor of #33

PR #34, opened 2016-05-21
Auto label PR's based on content

weaverryan:
Hi guys!

This finishes #29, after a few of us made big changes to the repository.

The functionality is best seen by looking at the `AutLabelPRFromContentSubscriberTest`.

This will only be used on `symfony/symfony`, not the docs.

I'm pretty confident this is solid. We just need approval that we for sure want this functionality :).

Ping @javiereguiluz


src/AppBundle/Subscriber/AutoLabelPRFromContentSubscriber.php
```diff
@@ -0,0 +1,110 @@
...
+
+    /**
+     * Adds a "Needs Review" label to new PRs.
+     *
+     * @param GitHubEvent $event
+     */
+    public function onPullRequest(GitHubEvent $event)
+    {
+        $data = $event->getData();
+        if ('opened' !== $action = $data['action']) {
```
HeahDude: Why not allow `edited` as suggested by @webmozart?

HeahDude:
Great work! I was doing the same :) That was fast!

src/AppBundle/Subscriber/AutoLabelPRFromContentSubscriber.php
```diff
@@ -0,0 +1,108 @@
...
+        $prBody = $data['pull_request']['body'];
+        $prLabels = array();
+
+        // the PR title usually contains one or more labels
+        foreach ($this->extractLabels($prTitle) as $label) {
+            $prLabels[] = $label;
+        }
+
+        // the PR body usually indicates if this is a Bug, Feature, BC Break or Deprecation
+        if (preg_match('/\|\s*Bug fix\?\s*\|\s*yes\s*/', $prBody, $matches)) {
```
WouterJ: let's use `/.../i` to allow "Yes" as well

src/AppBundle/Subscriber/AutoLabelPRFromContentSubscriber.php
```diff
@@ -0,0 +1,108 @@
...
+            'pull_request' => $prNumber,
+            'pr_labels' => $prLabels,
+        ));
+    }
+
+    private function extractLabels($prTitle)
+    {
+        $labels = array();
+
+        // e.g. "[PropertyAccess] [RFC] [WIP] Allow custom methods on property accesses"
```
WouterJ: what about automatically applying "Status: Needs Work" when the PR has the `[WIP]` label?

src/AppBundle/Subscriber/AutoLabelPRFromContentSubscriber.php
weaverryan: It's just a matter of adding it later - it could be really easy, i actually didn't check :)

javiereguiluz:
@weaverryan thanks for finishing this and adding proper tests. I agree with the minor comments made by @WouterJ, so this looks ready to me 👍 

Regarding your question: "do we want this functionality?"  Today we have this functionality enabled in the repo ... but it's made by hand instead by this bot. I think it's useful because Symfony's backlog is huge and these labels allow us to quickly filter by topic (bug/feature/deprecation) and by component.

weaverryan:
This is now active on the core repository! So, let's see how Carson performs :)

src/AppBundle/Subscriber/AutoLabelPRFromContentSubscriber.php
```diff
@@ -0,0 +1,117 @@
...
+        if (preg_match_all('/\[(?P<labels>.+)\]/U', $prTitle, $matches)) {
+            // creates a key=>val array, but the key is lowercased
+            $validLabels = array_combine(
+                array_map(function($s) {
```
sstok: This can be simplified to `array_map('strtolower`, $this->getValidLabels())`, no?

src/AppBundle/Subscriber/AutoLabelPRFromContentSubscriber.php
```diff
@@ -0,0 +1,117 @@
...
+                    return strtolower($s);
+                }, $this->getValidLabels()),
+                $this->getValidLabels()
+            );
+
+            foreach ($matches['labels'] as $label) {
+                // check case-insensitively, but the apply the correctly-cased label
+                if (isset($validLabels[strtolower($label)])) {
```
sstok: Mini optimization: `$label = strtolower($label);` before the if statement 👍

PR #36, opened 2016-05-21
[Docs] Auto-comment on new PR's with instructions

weaverryan:
Hi guys!

Whenever a new PR is made, carson will comment with a nice message and details about how anyone can change status. This is only for the docs.

Here is what the rough-draft of this message looks like: https://github.com/weaverryan/symfony/pull/3#issuecomment-220802253

Obviously, let's make sure it says _just_ want we want.

Thanks!


HeahDude:
Why did Carson add the "Status: works for me" and not "reviewed" which is the last in its comment?

weaverryan:
Great question! I have no idea :D

HeahDude:
Maybe we should add a "secret" escape option to the regex?

WouterJ:
I don't like this that much and mainly because of the length of the message. Imagine this being commented on *every* opened PR. It's like going back a few years and having Travis commenting each build status in a comment.

For new contributors, adding a comment about what's going to happen next ("Hey there, nice job! This is PR will now need to be reviewed [...]") might be usefull. But doing this for every PR doesn't seem great.

Ener-Getick:
@WouterJ i agree, maybe we could check if the CONTRIBUTORS.md file contains an occurrence of the username of the pr creator.

sstok:
I think it's safer to keep a "known" list of contributors in the application database (by user-id) rather then the username as this can change, and the CONTRIBUTORS.md file is not always up-to-date (open two pr's after each other for example).

weaverryan:
This is not only meant for new contributors, but also for reviewers to know that they are able to modify the status of the pull request.

Another option would be to dynamically edit the PR description to add a new, mini-section at the bottom with these details.

src/AppBundle/Subscriber/NewPRInstructionCommentSubscriber.php
```diff
@@ -0,0 +1,73 @@
...
+        if ('opened' !== $action = $data['action']) {
+            $event->setResponseData(array('unsupported_action' => $action));
+
+            return;
+        }
+
+        $needsReviewLabel = GitHubStatusApi::getNeedsReviewLabel();
+
+        $commentTexts = StatusChangeByCommentSubscriber::getValidCommentsForDisplay();
+        $commentBody = sprintf(<<<EOF
```
mickaelandrieu: Hi  @weaverryan, what do you think about use twig and configurable templates instead ?

Ener-Getick:
@weaverryan what do you think about including a link targeting the docs in the pr description ?
This way it wouldn't disturb regular contributors but it would help new comers.

Nyholm:
This is replaced by #95

PR #37, opened 2016-05-21
Handle PR edit

HeahDude:
What about this?

### DIFF
--- a/src/AppBundle/Subscriber/AutoLabelPRFromContentSubscriber.php
+++ b/src/AppBundle/Subscriber/AutoLabelPRFromContentSubscriber.php
@@ -25,7 +25,7 @@ class AutoLabelPRFromContentSubscriber implements EventSubscriberInterface
     public function onPullRequest(GitHubEvent $event)
     {
         $data = $event->getData();
-        if ('opened' !== $action = $data['action']) {
+        if (!in_array($action = $data['action'], array('opened', 'edited'), true)) {
             $event->setResponseData(array('unsupported_action' => $action));
 
             return;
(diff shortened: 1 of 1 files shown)


src/AppBundle/Subscriber/AutoLabelPRFromContentSubscriber.php
```diff
@@ -25,7 +25,7 @@ public function __construct(CachedLabelsApi $labelsApi)
     public function onPullRequest(GitHubEvent $event)
     {
         $data = $event->getData();
-        if ('opened' !== $action = $data['action']) {
+        if (!in_array($action = $data['action'], array('opened', 'edited'), true)) {
```
HeahDude: We could use a class to hold actions as constants like events. What do you think?

weaverryan: or a new value object for different types of events - e.g .`PullRequest` - with getters like `getAction()`, etc.

weaverryan:
@HeahDude can you add a test that tests the addition/removal of labels - that's the only complex part of this change, and I do *not* want Carson making mistakes :)

HeahDude:
@weaverryan Of courses, I thought about it and I'd love to! But I wanted your feedback first :)

Also, it's really hard to be sure about test I write since I cannot run them all... Could you please help me with this by adding some fixtures about an edited PR so I can work on that?

You could send a PR or push it so I can rebase, what do you think? Thanks!

src/AppBundle/Subscriber/AutoLabelPRFromContentSubscriber.php
HeahDude: Yeah I thought about that when originally making a subscriber factory. But I really think we should keep the data array (payload) and maybe use some specific events like an `GitHubIssueEvent` which can hold specific method like `getAction` and fallback on the generic event in most case.

This should be the goal of another PR, what do you think?

HeahDude:
I'm about to close this one, I've tested Carson with my own repo and it seems (the logs on github confirmed it) that we can't trust the payload concerning changes made to the title and description of any tickets :(

yceruto:
> we can't trust the payload concerning changes made to the title and description of any tickets :(

Some news about this ? 

* https://developer.github.com/v3/activity/events/types/#pullrequestevent
* https://developer.github.com/v3/activity/events/types/#issuesevent

HeahDude:
Thanks @yceruto for the doc, I'm aware of it ;) Sadly, AFAIK the payload is not safe regarding the change set (before/after), and in most on my real tests with my forks of Carson and projects the values in the payload was all about the old version.

I won't have time in the next weeks to look closer at this, feel free to test it on your side or take over this PR :)

sstok:
You could issue (no pun) a new API call to get the new state, by getting the actual issue/pull-request. It would be nice if GitHub could provide both the old and new state in the payload 🤔

HeahDude:
@sstok 

> You could issue (no pun) a new API call

I've thought about it but given the potential number of requests I've put that thought away.

> It would be nice if GitHub could provide both the old and new state in the payload

It does, it is just wrong most of the time with the new value of the changeset equal to the old one.

stof:
> It does, it is just wrong most of the time with the new value of the changeset equal to the old

@HeahDude Have you reported this to Github ? It looks like a bug.

PR #38, opened 2016-05-22
Switch "Needs Work" to "Needs Review" after a push

wouterj:


src/AppBundle/Subscriber/StatusChangeOnPushSubscriber.php
```diff
@@ -0,0 +1,54 @@
...
+ * commits.
+ *
+ * @author Wouter de Jong <wouter@wouterj.nl>
+ */
+class StatusChangeOnPushSubscriber implements EventSubscriberInterface
+{
+    public function onPullRequest(GitHubEvent $event)
+    {
+        $data = $event->getData();
+        if ('synchronize' === $data['action']) {
```
Ener-Getick: Shouldn't this be ``'edited' !== `` ?

WouterJ:
Added tests now and fixed some bugs. Should be ready if people like this change.

src/AppBundle/Subscriber/StatusChangeOnPushSubscriber.php
```diff
@@ -0,0 +1,61 @@
...
+
+            return;
+        }
+
+        $repository = $event->getRepository();
+        $issueNumber = $data['issue']['number'];
+        $responseData = array('issue' => $issueNumber);
+        $status = $this->statusApi->getIssueStatus($issueNumber, $repository);
+
+        if (Status::NEEDS_REVIEW === $status) {
```
Ener-Getick: do we really need this ?
imo it doesn't matter if carsonbot always call the github api

app/config/github.yml
```diff
@@ -23,12 +28,14 @@ parameters:
                 - app.subscriber.status_change_by_comment_subscriber
                 - app.subscriber.needs_review_new_pr_subscriber
                 - app.subscriber.bug_label_new_issue_subscriber
+                - app.subscriber.status_change_on_push_subscriber
```
weaverryan: I think we should just enable this on the docs for now to see how it works. We would need to get the core team to agree to this, and I really want to keep Carson from doing anything "annoying". On the docs, we're better guinea pigs :)

app/config/github.yml
```diff
@@ -23,12 +28,14 @@ parameters:
...
             # secret: change_me
         symfony/symfony-docs:
             subscribers:
                 - app.subscriber.status_change_by_comment_subscriber
                 - app.subscriber.needs_review_new_pr_subscriber
                 - app.subscriber.bug_label_new_issue_subscriber
+                - app.subscriber.status_change_on_push_subscriber
 
         # used in a functional test
         weaverryan/symfony:
```
weaverryan: Add it down here too. This is used in a functional test. We should also update the WebhookControllerTest to send a new, fake "push" event. I'm just *quite* keen on keeping very tight test coverage: I don't want Carson running around doing crazy things :)

src/AppBundle/Subscriber/StatusChangeOnPushSubscriber.php
```diff
@@ -0,0 +1,61 @@
...
+            $responseData['status_change'] = null;
+            $event->setResponseData($responseData);
+
+            return;
+        }
```
weaverryan: Perhaps we should only change the status if it's currently in the `NEEDS_WORK` status? For example, if something is already `REVIEWED`, it's possible this was a tiny fix that a maintainer asked for, and it should really stay in `REVIEWED`.

stof:
this should account for ``WIP`` in the title, in case #39 is merged first

WouterJ:
@stof I don't see how that affects this PR. On a push, a PR will be moved to Needs Work anyway (it doesn't care about the title).

stof:
This PR is about switching to "Needs review" on push, not to "Needs work".

app/config/github.yml
WouterJ: How do you generate those webhook_example requests?

WouterJ:
Oh, my bad @stof. I've now updated this PR to take into account all comments (except from the fake webhook, as I'm not sure how to do that)

app/config/github.yml
WouterJ: Very friendly ping @weaverryan (also, any news on enabling carson for the docs?)

HeahDude:
Not sure if we should change the status on push automatically, it could also be to reset the test suite. Commenting gives more control IMO and keeping the current behavior also prevents fixing non wanted Carson actions.

sstok:
Just a thought: When a pull-request is marked as reviewed/needs review you can compute a hash from the final patch (not commits as they can change with a force-push) and then check if this has changed with the push. If not then it was a simple rebase/test restart and the status can be left unchanged.

app/config/github.yml
wouterj: ping @weaverryan 😃

wouterj:
@HeahDude tests should be run locally. The exact reason behind this PR is that many people don't comment after pushing fix changes. This means that many PRs end up being in "Status: needs work" for way longer than needed.

HeahDude:
Now that #39 is closed, and given the changes you made about wip titles and status check I think this PR is nice, good work man :)

app/config/github.yml
```diff
@@ -34,13 +39,15 @@ parameters:
         symfony/symfony-docs:
             subscribers:
                 - app.subscriber.status_change_by_comment_subscriber
+                - app.subscriber.status_change_on_push_subscriber
```
weaverryan: Did you purposefully not add this to `symfony/symfony`?

wouterj: > I think we should just enable this on the docs for now to see how it works. We would need to get the core team to agree to this, and I really want to keep Carson from doing anything "annoying". On the docs, we're better guinea pigs :)

Quote from @weaverryan in https://github.com/carsonbot/carsonbot/pull/38#discussion_r64154578

😉

weaverryan:
I asked one question, but 👍 from me!

app/config/github.yml
weaverryan: ah crap. Well, let's do that

weaverryan:
Deployed! But not activated for `symfony/symfony` - we can add it a bit later, assuming we don't hate it on the docs

wouterj:
Cool, thanks!

PR #44, opened 2016-08-23
Enable auto label PR for the docs

wouterj:
We have component labels as well in the docs and people use `[Component]` tags too. Let's enable the subscriber :)


HeahDude:
Do the labels match https://github.com/carsonbot/carsonbot/blob/master/src/AppBundle/Subscriber/AutoLabelPRFromContentSubscriber.php#L94 in the docs too?

weaverryan:
Sorry for the MONTHS delay :)

PR #45, opened 2016-08-29
Change Uncofirmed/Confirmed labels into statuses

### LINKED ISSUE #43: Change Confirmed/Unconfirmed labels into statuses
To me, the "Confirmed" and "Unconfirmed" labels represent the two possible statuses of a bug, not a standalone kind of issue. 

That's why I propose to change them into "Status: Confirmed" and "Status: Unconfirmed", assignable **only on "Bug" labelled issues**.

More, I would like to see "Status: Unconfirmed" automatically assigned at the same time of the "Bug" one when creating an issue and don't see the "Status : Reviewed/Needs review" labels anymore for these very same issues (Bug).
For bugs we are talking about confirmation, not review, labels are already in place. 
For proposals/feature requests, talking of review doesn't feel adapted but we could still use it (for all issues that aren't bug reports).

WDYT?

chalasr:
Closes #43

### DIFF
--- a/src/AppBundle/Subscriber/BugLabelNewIssueSubscriber.php
+++ b/src/AppBundle/Subscriber/BugLabelNewIssueSubscriber.php
@@ -44,7 +44,7 @@ class BugLabelNewIssueSubscriber implements EventSubscriberInterface
             return;
         }
 
-        $newStatus = Status::NEEDS_REVIEW;
+        $newStatus = Status::UNCONFIRMED;
 
         $this->statusApi->setIssueStatus($issueNumber, $newStatus, $repository);
         $responseData['status_change'] = $newStatus;
--- a/src/AppBundle/Subscriber/StatusChangeByCommentSubscriber.php
+++ b/src/AppBundle/Subscriber/StatusChangeByCommentSubscriber.php
@@ -14,8 +14,10 @@ class StatusChangeByCommentSubscriber implements EventSubscriberInterface
     private static $triggerWordToStatus = [
         'needs review' => Status::NEEDS_REVIEW,
         'needs work' => Status::NEEDS_WORK,
+        'unconfirmed' => Status::UNCONFIRMED,
         'works for me' => Status::WORKS_FOR_ME,
         'reviewed' => Status::REVIEWED,
+        'confirmed' => Status::CONFIRMED,
     ];
 
     private $statusApi;
--- a/src/AppBundle/Issues/GitHub/GitHubStatusApi.php
+++ b/src/AppBundle/Issues/GitHub/GitHubStatusApi.php
@@ -12,8 +12,10 @@ class GitHubStatusApi implements StatusApi
     private static $statusToLabel = [
         Status::NEEDS_REVIEW => 'Status: Needs Review',
         Status::NEEDS_WORK => 'Status: Needs Work',
+        Status::UNCONFIRMED => 'Status: Unconfirmed',
         Status::WORKS_FOR_ME => 'Status: Works for me',
         Status::REVIEWED => 'Status: Reviewed',
+        Status::CONFIRMED => 'Status: Confirmed',
     ];
 
     private $labelToStatus = [];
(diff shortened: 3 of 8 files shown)


chalasr:
One doubt: Is it better to have `Status: Needs Confirmation` rather than `Status: Unconfirmed`? As the idea is to make it replaces the `Needs Review` for `Bug`' issues

src/AppBundle/Subscriber/BugLabelNewIssueSubscriber.php
```diff
@@ -36,15 +36,15 @@ public function onIssues(GitHubEvent $event)
...
+        // Ignore non-bugs or issue which already has a statu
         if ('bug' !== strtolower($data['label']['name']) || null !== $this->statusApi->getIssueStatus($issueNumber, $repository)) {
             $responseData['status_change'] = null;
             $event->setResponseData($responseData);
 
             return;
         }
 
-        $newStatus = Status::NEEDS_REVIEW;
+        $newStatus = Status::UNCONFIRMED;
```
HeahDude: Maybe both :)

src/AppBundle/Subscriber/BugLabelNewIssueSubscriber.php
```diff
@@ -36,15 +36,15 @@ public function onIssues(GitHubEvent $event)
         }
 
         $responseData = array('issue' => $issueNumber = $data['issue']['number']);
-        // Ignore non-bugs or issue which already has a status
+        // Ignore non-bugs or issue which already has a statu
```
HeahDude: please revert this change

src/AppBundle/Subscriber/StatusChangeByCommentSubscriber.php
```diff
@@ -14,8 +14,11 @@ class StatusChangeByCommentSubscriber implements EventSubscriberInterface
     private static $triggerWordToStatus = [
         'needs review' => Status::NEEDS_REVIEW,
         'needs work' => Status::NEEDS_WORK,
+        'needs work' => Status::NEEDS_WORK,
```
HeahDude: it seems you've duplicated this line

chalasr: oops... thank you

HeahDude:
Ok why not? But I think we should avoid mixing reviewed/confirmed concept and keep two distinct pair of statuses that can change to each other.

src/AppBundle/Subscriber/BugLabelNewIssueSubscriber.php
chalasr: Yeah I thought about this alternative :)
But as pointed in the ticket, IMHO issues don't need a review in general, more opinions, confirmations and maybe support. At least for Bugs, we are sure they just need confirmation, so maybe [this one](https://github.com/carsonbot/carsonbot/pull/45#issuecomment-243268636) can be a good compromise. more expressive imo, wdyt?

src/AppBundle/Subscriber/BugLabelNewIssueSubscriber.php
chalasr: Yeah I thought about this alternative :)
But as pointed in the ticket, IMHO issues don't need a review in general, more opinions, confirmations and maybe support. At least for Bugs, we are sure they just need confirmation, so maybe [this one](https://github.com/carsonbot/carsonbot/pull/45#issuecomment-243268636) can be a good compromise. more expressive imo, wdyt?

HeahDude: In case of a bug confirmed does not mean we have a valid solution for it and it may need some other reviews, on the contrary it might stay unconfirmed and set reviewed so we can close it.

chalasr: I didn't see neither the reviewed status used as "good to close" nor a confirmed bug closed due to lack of solutions (they stay open for ever ^^), but in this case I think you're right. 👍

chalasr:
@weaverryan Before making the changes according to @HeahDude proposal, could you please confirm we can envisage having issues with two statuses at the same time? 
"Needs review" and "Unconfirmed" would be auto assigned on Bugs, both could be changed distinctively like confirming the bug without reviewing it.

chalasr:
I just realized that the `Unconfirmed` label is removed when commenting `Status: reviewed`, I think it's good enough

PR #49, opened 2016-11-07
Consider "DI" an alias of "DependencyInjection" label

javiereguiluz:
I propose to fix just "DI" label because it's the only common case I know. If there were other cases, we should introduce a utility method to resolve these aliases.


javiereguiluz:
Closed in favor of #52.

PR #55, opened 2017-01-28
Added Inflector and WebServerBundle labels

wouterj:


chalasr:
What about adding DotEnv btw?

wouterj:
Sounds good, but `DotEnv` doesn't yet have a label in `symfony/symfony`, so it can't be in the valid label list.

sstok:
When the label does not exist GitHub registers it 😆 but only for the API btw.

xabbuh:
The `Dotenv` label is now present in both repositories symfony/symfony and symfony/symfony-docs.

wouterj:
Ready to merge

src/AppBundle/Subscriber/AutoLabelPRFromContentSubscriber.php
```diff
@@ -109,7 +109,7 @@ private function getValidLabels()
             'Asset', 'BC Break', 'BrowserKit', 'Bug', 'Cache', 'ClassLoader',
             'Config', 'Console', 'Critical', 'CssSelector', 'Debug', 'DebugBundle',
             'DependencyInjection', 'Deprecation', 'Doctrine', 'DoctrineBridge',
-            'DomCrawler', 'Drupal related', 'DX', 'Easy Pick', 'Enhancement',
+            'DomCrawler', 'DotEnv', 'Drupal related', 'DX', 'Easy Pick', 'Enhancement',
```
xabbuh: The name in fact is `Dotenv` (with a lowercased `e`).

HeahDude: This PR should be merged with the upcoming 3.3 release, and we also need "Image", "Lock" and "UTF-8" :).

chalasr: @HeahDude None of them for now :)

HeahDude: @chalasr Actually components and labels exist so Carson should be aware of them even if they are only in master to be able to do its job ;)

PR #56, opened 2017-01-28
Disabled status subscribers for the docs

wouterj:
As proposed in our Slack channel, it's probably better to use the GitHub review API instead of the status labels in the docs.

This PR disables setting the `Needs Review` label on creation and pushing a new commit. I still enabled status change by comment, as status labels still make sense for bug reports imo.

/cc @javiereguiluz @weaverryan @xabbuh

### DIFF
--- a/app/config/github.yml
+++ b/app/config/github.yml
@@ -40,8 +40,8 @@ parameters:
         symfony/symfony-docs:
             subscribers:
                 - app.subscriber.status_change_by_comment_subscriber
-                - app.subscriber.status_change_on_push_subscriber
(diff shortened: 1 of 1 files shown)


javiereguiluz:
👍 Thanks!

Status: reviewed

xabbuh:
:+1:

stof:
I think #58 should be done first IMO

stof:
And moving it back to Needs review when doing new changes in the PR is not something you can do with github reviews

stof:
actually, I think this PR should not be merged at all (but #58 should be)

PR #57, opened 2017-02-01
Do not enable label if PR body line is equal to "yes/no"

chalasr:
Sometimes PR body isn't filled, it happened to me by mistake already (hit enter) and, when it isn't, all lines  equal to `yes/no` are interpreted as `yes` by carson , the result being that all possible labels are added to the PR and then one has to ask for relabelling.
See [remaining examples](https://github.com/symfony/symfony/pulls?utf8=%E2%9C%93&q=is%3Apr%20label%3AFeature%20label%3ABug%20label%3A%22BC%20Break%22%20label%3ADeprecation%20).


nicolas-grekas:
Having labels become like a Christmas tree may be a nice thing in fact: it clearly encourages the submitter to do better next time.

Issue #50, opened 2016-11-09
Review summaries are ignored


ogizanagi:
It seems that the review summaries introduced by Github in ["A whole new Github universe: Code better with reviews"](https://github.com/blog/2256-a-whole-new-github-universe-announcing-new-tools-forums-and-features#code-better-with-reviews) are ignored by carsonbot.

See https://github.com/symfony/symfony/pull/20418#pullrequestreview-7846023 for instance.

xabbuh:
Note that we will have to wait for GitHub to expose review comments one way or the other through their API so that the KnpLabs library can be updated if needed.

yceruto:
@ogizanagi : In the meantime, could you change the status to reviewed using the traditional comment? :smile: Thank you!

ogizanagi:
@yceruto : Now that the issue is reported, for sure! 😅

PR #58, opened 2017-03-07
Added support for GitHub reviews

wouterj:
Currently, `status:` and such in a review body isn't parsed by Carson. This is a bit annoying, as it means we either have to manually assign the status label or write an issue comment with just the `status:`.

This PR adds a new status subscriber for GitHub reviews. Besides parsing review bodies, this subscriber does 2 more things:

 * Updating the status based on the review state (approved/changes requested)
 * Updating the status to `needs review` when a GitHub review is requested

@weaverryan I'm still not sure how to implement the functional testing bit, if you can do that before/after the merge, I would be very happy :)


src/AppBundle/Subscriber/StatusChangeByReviewSubscriber.php
```diff
@@ -0,0 +1,115 @@
...
+
+            return;
+        }
+
+        $repository = $event->getRepository();
+        $pullRequestNumber = $data['pull_request']['number'];
+        $newStatus = null;
+
+        // Set status based on review state
+        switch (strtolower($data['review']['state'])) {
```
wouterj: GitHub docs are a bit vague on this: REST api shows state in all uppercase, while the webhooks show it in lowercase. Just to be sure, I've added this `strtolower()` call.

src/AppBundle/Subscriber/StatusChangeByReviewSubscriber.php
```diff
@@ -0,0 +1,115 @@
...
+                break;
+            default:
+                $newStatus = $this->parseStatusFromText($data['review']['body']);
+
+                if (Status::REVIEWED === $newStatus && false === $this->checkUserIsAllowedToReview($data)) {
+                    $newStatus = null;
+                }
+        }
+
+        $event->setResponseData(array(
```
HeahDude: What about moving the beginning of the method and those line in the abstract listener too, it could call an abstract method to extract the status from the payload, wdyt?

src/AppBundle/Subscriber/AbstractStatusChangeSubscriber.php
```diff
@@ -0,0 +1,52 @@
...
+     * status change.
+     *
+     * @param string $body
+     *
+     * @return null|string
+     */
+    protected function parseStatusFromText($body)
+    {
+        $triggerWord = implode('|', array_keys(static::$triggerWordToStatus));
+        $formatting = '[\\s\\*]*';
```
sstok: FYI you don't have to escape special characters in a character class, except `-` if it's not at the end and to be used as literal.

src/AppBundle/Subscriber/StatusChangeByReviewSubscriber.php
```diff
@@ -0,0 +1,115 @@
...
+
+    public static function getSubscribedEvents()
+    {
+        return array(
+            GitHubEvents::PULL_REQUEST_REVIEW => 'onReview',
+            GitHubEvents::PULL_REQUEST => 'onReviewRequested',
+        );
+    }
+
+    private function checkUserIsAllowedToReview(array $data)
```
sstok: This method is also defined in AbstractStatusChangeSubscriber, and check implies it will throw exception or something. By Symfony conventions it should be `isUserIsAllowedToReview` 👍

src/AppBundle/Tests/Subscriber/StatusChangeByReviewTest.php
```diff
@@ -0,0 +1,149 @@
...
+    private $repository;
+
+    /**
+     * @var EventDispatcher
+     */
+    private static $dispatcher;
+
+    public static function setUpBeforeClass()
+    {
+        self::$dispatcher = new EventDispatcher();
```
stof: this should not be a static property. The dispatcher instance must not be shared between tests, otherwise the listener will run several times for subsequent tests (as listeners registered by previous tests will still be here).
I faced this issue when implementing the same feature before seeing your PR, because I added a check in the subscriber to avoid changing the status if it is already the current one. (making the test fail when a previous listener was running too with the config of the previous test).

stof:
@weaverryan any news on it ?

stof:
@weaverryan ping

weaverryan:
I just made some changes! I would appreciate any review - it's still a bit of "guess work" when processing webhooks, and I'd like to not get it wrong :)

stof:
this PR should close both #50 and #59

weaverryan:
This is merged and deployed! But, I don't know if the `Pull request review` webhook is checked on `symfony/symfony` or not. So, if this doesn't work, we need to ping fabpot about that. I just tested on `weaverryan/symfony` and the feature definitely works (assuming that webhook is being delivered)

fabpot:
Pull request reviews was already checked on symfony/symfony, and I've just checked it on symfony/symfony-docs.

Jean85:
Can confirm that it's working now! :+1: https://github.com/symfony/symfony-docs/pull/8899#pullrequestreview-117854022

PR #68, opened 2019-03-28
Add feature to create issues on symfony/symfony-docs

Nyholm:
It would be nice if we always created an issue on Symfony/symfony-docs when we merge a feature in Symfony/symfony. 

This will write an issue on only if:
- The PR is merged
- The PR has label "Feature"
- The PR description/body does not contain "symfony/symfony-docs"

### DIFF
--- a/app/config/github.yml
+++ b/app/config/github.yml
@@ -32,8 +32,8 @@ services:
             - '@app.status_api'
             - '@logger'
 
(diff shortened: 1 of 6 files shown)


weaverryan:
Thanks for the PR - but apparently this is already happening via the internal merge tool - e.g. https://github.com/symfony/symfony-docs/issues/11850

PR #79, opened 2019-10-02
Get valid labels from the repository via GitHub API

yceruto:
https://github.com/carsonbot/carsonbot/blob/686b5b2298f98487c59ea7b29a6240cb3389ab46/src/AppBundle/Subscriber/AutoLabelPRFromContentSubscriber.php#L103-L106

This is the most common change in this repo :) so let's make it automatic.

/cc @weaverryan


wouterj:
Ping @weaverryan 

If we don't merge this one, we should add the Notifier label. Yet, I think it makes most sense to make this dynamic.

PR #81, opened 2019-10-04
Upgrade to 4.x

j4nr6n:
This is a HUGE WIP!

* [ ] Fix tests

@weaverryan I want to apologize in advance for so many changed files.


j4nr6n:
Got the rest of the tests added. Still two failures, but now there are a bunch of errors for things like type mismatches and such.

config/services.yaml
```diff
@@ -0,0 +1,121 @@
...
+        class: App\Subscriber\BugLabelNewIssueSubscriber
+        arguments:
+            - '@App\Issues\GitHub\GitHubStatusApi'
+        public: true
+
+    app.subscriber.auto_label_pr_from_content_subscriber:
+        class: App\Subscriber\AutoLabelPRFromContentSubscriber
+        arguments:
+            - '@App\Issues\GitHub\CachedLabelsApi'
+        public: true
```
j4nr6n: Not sure why I had to make all of these public. I tried to implement `ServiceSubscriberInterface` in `GitHubRequestHandler` but I couldn't get it to work.

.symfony.cloud.yaml
```diff
@@ -6,44 +6,41 @@ runtime:
...
     deploy: |
         set -x -e
 
         # "install" cache
         # If your cache can be read-only, you can skip these steps
-        rm -rf app/cache/prod
-        #cp -Rp tmp/bootstrap.php.cache app/bootstrap.php.cache
-        cp -Rp tmp/cache/prod app/cache/
+        rm -rf var/cache/prod
+        cp -Rp tmp/cache/prod var/cache/
```
tucksaun: You need to use this snippet instead: https://github.com/symfonycorp/cloud-templates/blob/master/01-silex.yaml#L34-L47

config/services.yaml
```diff
@@ -0,0 +1,121 @@
...
+    # add more service definitions when explicit configuration is needed
+    # please note that last definitions always *replace* previous ones
+
+    Github\Client:
+        arguments: []
+        calls:
+            - method: authenticate
+              arguments:
+                  - http_token
+                  - '%env(GITHUB_TOKEN)%'
```
tucksaun: I think this should be moved (and replace) the definition in `config/packages/github_api.yaml`

.env
```diff
@@ -0,0 +1,32 @@
...
+APP_ENV=dev
+APP_SECRET=172e34116290b1919eff99c26ef17dca
+#TRUSTED_PROXIES=127.0.0.1,127.0.0.2
+#TRUSTED_HOSTS='^localhost|example\.com$'
+###< symfony/framework-bundle ###
+
+###> knplabs/github-api ###
+GITHUB_AUTH_METHOD=http_password
+GITHUB_USERNAME=username
+GITHUB_SECRET=password_or_token
```
tucksaun: seem unused

.env
```diff
@@ -0,0 +1,32 @@
...
+
+###> symfony/framework-bundle ###
+APP_ENV=dev
+APP_SECRET=172e34116290b1919eff99c26ef17dca
+#TRUSTED_PROXIES=127.0.0.1,127.0.0.2
+#TRUSTED_HOSTS='^localhost|example\.com$'
+###< symfony/framework-bundle ###
+
+###> knplabs/github-api ###
+GITHUB_AUTH_METHOD=http_password
```
tucksaun:
```suggestion
GITHUB_AUTH_METHOD=http_token
```

.env
j4nr6n: They were added as part of a bundle's recipe.

.env
j4nr6n: I'll look into these a bit more. They were part of a bundle's recipe and not used in the old app so I didn't think to check them.

config/services.yaml
j4nr6n: Thanks. I'll look into that.

.symfony.cloud.yaml
j4nr6n: I'm not familiar with what is going on in this file. I limited my changes to obvious ones (like changing `app/cache` to `var/cache`, or removing the front-controller since there isn't a "dev" front-controller anymore). I'm not sure I'm the right person to make changes to the existing behavior. Especially considering I can't test those changes.

PR #84, opened 2020-04-03
fix(auto-label): resynchronize valid labels

fancyweb:
I noticed some labels were not added by carsonbot on symfony/symfony.

I resynchronized the valid labels here (cleaned and added some).


weaverryan:
Thank you @fancyweb!

PR #85, opened 2020-06-17
update to Symfony 5.1

IonBazan:
This PR bumps all packages to most recent versions. Unused Symfony components have been dropped (like Security). Doctrine has been dropped too as it seems unused.
@weaverryan please test it on your Github account when you have a while to check if it's working properly.


.env
```diff
@@ -0,0 +1,34 @@
...
+#
+# Run "composer dump-env prod" to compile .env files for production use (requires symfony/flex >=1.2).
+# https://symfony.com/doc/current/best_practices.html#use-environment-variables-for-infrastructure-configuration
+
+###> symfony/swiftmailer-bundle ###
+# For Gmail as a transport, use: "gmail://username:password@localhost"
+# For a generic SMTP server, use: "smtp://localhost:25?encryption=&auth_mode="
+# Delivery is disabled by default via "null://localhost"
+MAILER_URL=null://localhost
+###< symfony/swiftmailer-bundle ###
```
weaverryan, line 21: We can remove swiftmailer (and all the related config files, should be done automatically when you remove it via the recipe)

config/github.yml
```diff
@@ -1,36 +1,31 @@
...
+        class: App\Subscriber\AutoLabelPRFromContentSubscriber
+        public: true
 
     app.subscriber.status_change_by_review_subscriber:
-        class: AppBundle\Subscriber\StatusChangeByReviewSubscriber
-        arguments:
-            - '@app.status_api'
-            - '@logger'
+        class: App\Subscriber\StatusChangeByReviewSubscriber
+        public: true
```
weaverryan, line 28: Could we remove this file entirely? And fully register everything as a class name (i.e.remove the snake case ids). I'd rather fully modernize the code base (and services) now, test it, and be done :)

config/services.yaml
```diff
@@ -0,0 +1,54 @@
...
+    # default configuration for services in *this* file
+    _defaults:
+        autowire: true      # Automatically injects dependencies in your services.
+        autoconfigure: true # Automatically registers your services as commands, event subscribers, etc.
+
+    # makes classes in src/ available to be used as services
+    # this creates a service per class whose id is the fully-qualified class name
+    App\:
+        resource: '../src/*'
+        exclude: '../src/{DependencyInjection,Subscriber,Issues,Entity,Migrations,Event,Tests,Kernel.php,GitHubEvents.php}'
```
weaverryan, line 22: No reason to over-optimize the exclude:

* I think Subscriber can be removed if we remove `github.yml` and allow those to be auto-registered like normal
* I'd remove `Issues`. Sure, some things are model classes, but it doesn't hurt to include them. Plus, then it would auto-register `GitHubRequestHandler` and the 2 services in the `Github/` sub-directory

config/services.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+    Github\Api\Issue\Labels:
+        factory: ['@Github\Api\Issue', labels]
+
+    App\Issues\GitHub\CachedLabelsApi: ~
+
+    App\Issues\GitHub\GitHubStatusApi: ~
+
+    App\Issues\StatusApi:
+        alias: App\Issues\GitHub\GitHubStatusApi
```
weaverryan, line 45: These can shorten to

```yml
App\Issues\StatusApi: '@ App\Issues\GitHub\GitHubStatusApi'
```

config/services.yaml
weaverryan: Later

config/github.yml
weaverryan: Later

weaverryan:
Thank you Ion!

PR #86, opened 2020-07-07
refactor services to be private and build configuration at build time

IonBazan:


config/github.yml
```diff
@@ -1,60 +1,41 @@
...
-
-    app.subscriber.status_change_by_review_subscriber:
-        class: App\Subscriber\StatusChangeByReviewSubscriber
-        public: true
+    App\Subscriber\:
+        resource: '../src/Subscriber/*'
 
 parameters:
+    env(SYMFONY_DOCS_SECRET): ''
+    env(SYMFONY_SECRET): ''
```
weaverryan, line 11: Since we have these both defined in `.env`, I don't think we also need to give them defaults here

config/github.yml
```diff
@@ -1,60 +1,41 @@
...
-
-    app.subscriber.auto_label_pr_from_content_subscriber:
-        class: App\Subscriber\AutoLabelPRFromContentSubscriber
-        public: true
-
-    app.subscriber.status_change_by_review_subscriber:
-        class: App\Subscriber\StatusChangeByReviewSubscriber
-        public: true
+    App\Subscriber\:
+        resource: '../src/Subscriber/*'
```
weaverryan, line 7: Let's just move this whole file into `services.yaml` for simplicity. You can add the `autoconfigure: false` right after `resource` to disable it for just this set of services.

src/Repository/Repository.php
```diff
@@ -80,8 +58,13 @@ public function getNeedsReviewUrl()
...
-    public function getFullName()
+    public function getFullName(): string
     {
         return sprintf('%s/%s', $this->getVendor(), $this->getName());
     }
+
+    public function __toString(): string
+    {
+        return $this->getFullName();
+    }
```
weaverryan, line 69: If we can, I'd rather not have to rely on a `__toString` - i.e. let's explicitly call `getFullName()` wherever we're using this.

IonBazan:
@weaverryan Thanks for the review. I've applied suggested changes. Dependencies bumped too.

weaverryan:
Thanks for the nice PR - sorry for the delay Ion!

PR #87, opened 2020-10-26
Add support for automatically add milestone

Nyholm:
This will add a milestone to a PR if all the following conditions are true: 

- The PR is not targeting the default branch
- The PR target has the same name as a milestone

If we always added a milestone, it would be hard to know the difference between "A milestone is selected properly" and "Milestone selected but might be wrong". So this will be forgiving for contributors to submit a PR to 5.x (default branch) but the PR was intended for 4.4. Ie, maintainers can be sure that if a milestone is set, it is correct.


tests/Controller/WebhookControllerTest.php
```diff
@@ -58,6 +58,11 @@ public function getTests()
                 'pull_request.opened.json',
                 ['pull_request' => 3, 'status_change' => 'needs_review', 'pr_labels' => ['Bug']],
             ],
+            'On pull request opened with target branch' => [
+                'pull_request',
+                'pull_request.opened_target_branch.json',
+                ['pull_request' => 5, 'status_change' => 'needs_review', 'pr_labels' => ['Bug'], 'milestone'=>'4.4'],
```
Nyholm, line 64: How can I test this properly?

src/Subscriber/MilestoneNewPRSubscriber.php
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * Sets milestone on PRs that target non-default branch.
+     *
+     * @param GitHubEvent $event
+     */
+    public function onPullRequest(GitHubEvent $event)
+    {
+        $data = $event->getData();
+        $repository = $event->getRepository();
+        if ('opened' !== $action = $data['action']) {
```
weaverryan, line 34: `$action` variable not needed here

src/Issues/GitHub/MilestonesApi.php
```diff
@@ -0,0 +1,88 @@
...
+
+        $this->issuesApi->update($repository->getVendor(), $repository->getName(), $issueNumber, [
+            'milestone' => $milestoneNumber,
+        ]);
+    }
+
+    /**
+     * @return bool
+     */
+    public function exits(Repository $repository, string $milestoneName)
```
weaverryan, line 73: I think this should be `exists()`. You actually consistently named it, so I think the code *would* work as is ;)

tests/Controller/WebhookControllerTest.php
weaverryan: Not really a great way - it's a very basic "smoke test" :/

weaverryan (commented):
Minor tweaks!

src/Issues/GitHub/MilestonesApi.php
Nyholm: oops =)

src/Subscriber/MilestoneNewPRSubscriber.php
Nyholm: Thank you

Nyholm:
Thank you for the review. I've updated the RP

Nyholm:
Thank you for merging

PR #90, opened 2020-11-01
Added some simple CI

Nyholm:


composer.json
```diff
@@ -9,13 +9,14 @@
         "nyholm/psr7": "^1.3",
         "sensio/framework-extra-bundle": "^5.1",
         "symfony/console": "^5.1",
+        "symfony/dotenv": "^5.1",
         "symfony/flex": "^1.8",
         "symfony/framework-bundle": "^5.1",
         "symfony/http-client": "^5.1",
         "symfony/monolog-bundle": "~3.5",
+        "symfony/security-core": "5.1.*",
```
Nyholm, line 17: Here is the only "real" change. We were using classes from this package without having it installed

Nyholm:
Im done with this PR now.

CI will not run until merged. The CI will fail on master, there are some CS fixes and psalm issues that should be taken care of. 
Also the functional tests will fail.


I suggest merging this PR now and I'll work on making CI green in future PRs.

.github/workflows/static.yml
```diff
@@ -0,0 +1,57 @@
...
+jobs:
+  phpstan:
+    name: PHPStan
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Set up PHP
+        uses: shivammathur/setup-php@2.7.0
+        with:
+          php-version: '7.4'
```
Nyholm, line 18: Let's change this to whatever version we run on the production server

weaverryan: 7.2 at the moment - but we should maybe keep this and then upgrade to 7.4 https://github.com/carsonbot/carsonbot/blob/master/.symfony.cloud.yaml

Nyholm: Okey. Cool cool cool

Nyholm:
Thank you for merging

PR #91, opened 2020-11-01
Playground request data

Nyholm:
Here is a big PR that prepare us to run functional tests on CI. 

Blocked by #90


config/packages/github_api.yaml
```diff
@@ -3,7 +3,7 @@ services:
         arguments:
             - '@Github\HttpClient\Builder'
         calls:
-            - ['authenticate', ['%env(GITHUB_TOKEN)%', 'http_token']]
+            - ['authenticate', ['%env(GITHUB_TOKEN)%', null, 'access_token_header']]
```
Nyholm, line 6: Note, this will break production. 

The `carsonbot` user needs to login to github> Go to settings > developers settings > personal access token (link: https://github.com/settings/tokens) and then generate a new access token with "api" privileges.

Nyholm:
The badge is useful because it will show if the integration tests on master is passing or not

config/packages/github_api.yaml
weaverryan: Can you explain this a bit more? carsonbot already has a personal access token, with with just one scope - `public_repo`. What difference does this change make? And how would this new API token be different? I don't see anything specific to "api" privileges when I create a token.

Makefile
```diff
@@ -0,0 +1,12 @@
...
+help:
+	@grep -E '^[a-zA-Z0-9_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
+
+build: test ## Runs test targets
+
+test: vendor/autoload.php ## Runs tests with phpunit
+	vendor/bin/simple-phpunit --testsuite Unit
+
+vendor/autoload.php:
+	composer install --no-interaction
```
weaverryan, line 12: Is this file used?

Nyholm: When running `make test`, this make sure to run `composer install` if the `vendor/autoload.php` does not exist. 

So yes.

config/packages/github_api.yaml
Nyholm: I am wrong. 

The `http_token` and `access_token_header` is the same. But the former is a deprecated value. 
https://github.com/KnpLabs/php-github-api/blob/2.x/lib/Github/HttpClient/Plugin/Authentication.php#L54-L58

No bc break, this change is fine.

weaverryan:
Thanks Tobias!

PR #94, opened 2020-11-01
Add a custom retry strategy to retry 404

Nyholm:
### Problem
Sometimes, we get a webhook request and then we trying to ask the Github api for that issue, but the API says 404.

### Acknowledgement

I know I do more service definitions than needed, but I didnt want to pull in FrameworkBundle:5.2

### Excuse 

The changes in src/Issues/GitHub/GitHubStatusApi.php is just CS.. I started to read/think there before I ended up with the RetryStrategy solution

### Shout out

@jderusse, this is really awesome =)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
         "symfony/dotenv": "^5.1",
         "symfony/flex": "^1.8",
         "symfony/framework-bundle": "^5.1",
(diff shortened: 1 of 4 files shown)


weaverryan:
Thanks again!

PR #96, opened 2020-11-02
Make PHPStan happy

Nyholm:


Nyholm:
It is phpstan 0.12.50 that will complain. We are running 0.12.48 =)

PR #97, opened 2020-11-02
Update PR title on label change

### LINKED ISSUE #9: Fix PR title prefixes
Once #8 is implemented, the bot should fix the title prefixes of PRs whenever their component/bundle/bridge labels are changed:
- `[<component>]` prefixes should be added for each component/bundle/bridge label
- superfluous `[xxx]` prefixes (e.g. `[WIP]`) should be removed

The idea is to retain the component labels of a PR in the Git log.

Nyholm:
This will fix #9 

Whenever a label is added, we make sure the title has the correct prefix.


src/Service/LabelNameExtractor.php
```diff
@@ -0,0 +1,77 @@
...
+class LabelNameExtractor
+{
+    private $labelsApi;
+
+    private static $labelAliases = [
+        'di' => 'DependencyInjection',
+        'bridge\twig' => 'TwigBridge',
+        'router' => 'Routing',
+        'translation' => 'Translator',
+        'twig bridge' => 'TwigBridge',
```
nicolas-grekas, line 22: is this frequent enough?
for bridges, all of them can be FooBridge or Bridge[\/]Foo, maybe we should have some logic for them?

src/Service/LabelNameExtractor.php
```diff
@@ -0,0 +1,77 @@
...
+        'wdt' => 'WebProfilerBundle',
```
nicolas-grekas, line 23: add fwb for FrameworkBundle

Nyholm: sure

src/Service/LabelNameExtractor.php
Nyholm: Im not sure..
This list is super old. I just moved it. 

I'll add other bridges too

Nyholm:
Question: Do we want "[Bug]" and "[BC Break]" in the PR title?

Nyholm:
No, we dont want that, and we dont get that. 

Everything is fine =)

src/Subscriber/AutoUpdateTitleWithLabelSubscriber.php
```diff
@@ -0,0 +1,79 @@
...
+            return;
+        }
+
+        $originalTitle = $prTitle = $data['pull_request']['title'];
+        $validLabels = [];
+        foreach ($data['pull_request']['labels'] as $label) {
+            if ('dddddd' === $label['color']) {
+                $validLabels[] = $label['name'];
+                // Remove label name from title
+                $prTitle = str_replace('['.$label['name'].']', '', $prTitle);
```
Nyholm, line 48: If the title is `[fwb] Foo`, then this line will check for `[FrameworkBundle]`, that will not be found and the final title will be: 

`[FrameworkBundle] [fwb] Foo`

Maybe we should remove all `[.*]` in the beginning of the title?

Nyholm: This is fixed and I also added a test for it.

wouterj:
I noticed a small issue with this change: https://github.com/symfony/symfony/pull/39036#event-3970838217 It seems like any label change (including the ones done by Carsonbot :smile: ) trigger changing the title. This results in the quite mysterious: "carsonbot changed the title <s>Add Romanian missing translations</s> Add Romanian missing translations"

Nyholm:
Lovely. 
I appreciate this bug report.

PR #102, opened 2020-11-03
Adding support for scheduled tasks

Nyholm:
This is awfully similar to Messenger component. 

I decided on this approach because I might want to schedule a task for 2-3 weeks. That is too much for Messenger's `DelayStamp` with AMQP.

### DIFF
--- a/src/Entity/Task.php
+++ b/src/Entity/Task.php
@@ -1,0 +1,114 @@
+<?php
+
+namespace App\Entity;
+
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * A "Task" is a scheduled job.
+ *
+ * @ORM\Table
+ * @ORM\Entity(repositoryClass="App\Repository\TaskRepository")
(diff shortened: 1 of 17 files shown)


src/Entity/Task.php
```diff
@@ -0,0 +1,114 @@
...
+ * @ORM\Table
+ * @ORM\Entity(repositoryClass="App\Repository\TaskRepository")
+ * @ORM\HasLifecycleCallbacks()
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class Task
+{
+    const ACTION_CLOSE_STALE = 1;
+    const ACTION_CLOSE_DRAFT = 2;
```
weaverryan, line 19: it's ok, but these aren't used in this PR

Nyholm: Correct. This is the teaser for upcoming PRs.

Nyholm:
Thank you for merging

PR #103, opened 2020-11-04
Add github workflow to verify docs

Nyholm:
@javiereguiluz and @wouterj may have noticed that I found a few invalid use statement recently. I was trying out my script. 

This PR will make sure that Carson is opening an issue when there are invalid use statements. 

TODO: 
- [x] Write logic to open/update an issue. (#106)

I'll wait for other PRs to be merged to avoid conflicts. But I just wanted to show my intent.

### DIFF
--- a/.github/workflows/docs-invalid-use.yml
+++ b/.github/workflows/docs-invalid-use.yml
@@ -1,0 +1,70 @@
+name: Verify docs
+
+on:
+  schedule:
+    - cron: '58 6 * * *'
+
+jobs:
+  use:
+    name: Invalid use statements
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+
+    steps:
+      - name: Set up PHP
+        uses: shivammathur/setup-php@2.7.0
+        with:
+          php-version: 7.4
+          coverage: none
+
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Checkout Symfony repo
+        run: git clone https://github.com/symfony/symfony .github/workflows/docs-invalid-use/symfony
+
+      - name: Checkout Symfony Docs repo
+        run: git clone https://github.com/symfony/symfony-docs .github/workflows/docs-invalid-use/docs
+
+      - name: Get composer cache directory
+        id: composer-cache
+        run: echo "::set-output name=dir::$(composer config cache-files-dir)"
+
+      - name: Cache dependencies
+        uses: actions/cache@v2
+        with:
+          path: ${{ steps.composer-cache.outputs.dir }}
+          key: composer-${{ runner.os }}-7.4-${{ hashFiles('composer.*') }}
+          restore-keys: |
+            composer-${{ runner.os }}-7.4-
+            composer-${{ runner.os }}-
+            composer-
+
+      - name: Download dependencies
+        run: |
+          composer install --no-interaction --optimize-autoloader
+          cd .github/workflows/docs-invalid-use
+          composer update --prefer-stable --no-interaction --optimize-autoloader
+          cd symfony
+          composer update --prefer-stable --no-interaction --optimize-autoloader
+
+      - name: Verify docs
+        run: |
+          cd .github/workflows/docs-invalid-use
+          ./run.php `pwd`/docs `pwd`/symfony > output.txt
+          cat output.txt
+
+      - name: Open issue
+        env:
+          GITHUB_TOKEN: ${{ secrets.CARSONPROD_GITHUB_TOKEN }}
+        run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/docs-invalid-use/run.php
```diff
@@ -0,0 +1,92 @@
...
+        }
+    }
+}
+
+if ($count === 0) {
+    error_log("We found nothing\n");
+}
+
+echo "foo".PHP_EOL;
+echo "bar".PHP_EOL;
```
Nyholm, line 71: This is just debug code

.github/workflows/docs-invalid-use.yml
```diff
@@ -0,0 +1,75 @@
+name: Verify docs
+
+on:
+  schedule:
+    - cron: '58 6 * * *'
+  # This is just for test
+  pull_request: ~
+  push:
+    branches:
+      - master
```
Nyholm, line 10: This is also just debug code

.github/workflows/docs-invalid-use.yml
```diff
@@ -0,0 +1,75 @@
...
+
+      - name: Verify docs
+        run: |
+          cd .github/workflows/docs-invalid-use
+          ./run.php `pwd`/docs `pwd`/symfony > output.txt
+          cat output.txt
+
+      - name: Open issue
+        env:
+          GITHUB_TOKEN: ${{ secrets.CARSONPROD_GITHUB_TOKEN }}
```
Nyholm, line 65: This needs to be added as a secret

wouterj:
Hi @Nyholm. I like your excitement for improving Carsonbot!

However, I think it's better to add this rule to https://github.com/OskarStark/doctor-rst . It's the CI tool we're using the docs repository to check for all sorts of things (including PHP stuff, like alphabetically sorted use-statements). This way, we're immediately reminded of the error when it's introduced.

What do you think?

Nyholm:
That is an excellent question. I’ve considered that too. But I decided not to do that. 
Most of the invalid use statements in the docs is because the symfony source changed. Ie, the docs were valid at the time of merge. 

If we added the check to the docs CI that would lead to the next PR made to the docs had red CI because unrelated changes. 

This test is also relatively slow and most of the time it will not report any issues.

wouterj:
I honestly am still not very sure about including this in carsonbot. I think the scope of carsonbot is management of PRs and issues (labelling, milestoning, commenting, etc). The scope of PhpUnit and DOCtor-rst is to validate the repository sources.

I agree that it might not be nice to break someones pull requests because of a change not made by them. We could make this job branch-only (so pull requests don't break). Also, classes in Symfony aren't suddenly moved. That only occurs on major releases and have a deprecation before. That means that there has been a doc issue about the move already, before the build actually breaks.

Nyholm:
I would like to extend that scope. I would like Carson bot to also open issues like: https://github.com/symfony/symfony/issues/38771 Currently it is another bot/script that opens that issue. 

I also think this is not going to happen very often. It is only when we introduce new big features and refactor them. Ie, the RateLimit component was refactored a bit after the docs got merged. 

At the very most, you will have one issue open on symfony/symfony-docs. And I dont really expect the bot to open more than 1 or 2 issues in 2020. 

The reason why I think this is important is because it is really really hard to verify this manually.

.github/workflows/docs-invalid-use/run.php
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/env php
+<?php
+
+if ($argc !== 3) {
+    echo "./docs-invalid-use.php path-to-docs path-to-symfony\n";
```
Nyholm, line 5: Typo

.github/workflows/docs-invalid-use/run.php
```diff
@@ -0,0 +1,89 @@
...
+    if (!file_exists($autoloadPHP)) {
+        echo "File $autoloadPHP does not exist.\nMake sure to run 'composer update'\n";
+        exit(1);
+    }
+    require $autoloadPHP;
+}
+
+use Symfony\Component\Finder\Finder;
+$finder = new Finder();
+$finder->in($docs)->name('*.rst');
```
OskarStark, line 28: You need to include rst.inc files too

stof:
@Nyholm I think this would be better handled by a workflow running in the symfony-docs repo (based on a cron trigger) that would use https://github.com/symfony-tools/code-block-checker

Nyholm:
Im closing this in favour of https://github.com/symfony-tools/code-block-checker/pull/38

That would make this feature part of the CI instead of Carson.

PR #109, opened 2020-11-07
Comment on stale issues

Nyholm:
This PR adds a `PingStaleIssuesCommand`. It will look for old inactive issues and start a process with them. 
1. Bot will make a comment to encourage activity and add label "Staled". 
2. Bot will make a comment to inform the issue will be closed
3. Bot will close the issue. 
The process can be interrupted with anyone making a comment on the issue or the "Keep open" label is added. 

The exact times between the steps described above is defined as constants in `PingStaleIssuesCommand`.

### DIFF
--- a/src/Api/Issue/GithubIssueApi.php
+++ b/src/Api/Issue/GithubIssueApi.php
@@ -31,8 +31,8 @@ class GithubIssueApi implements IssueApi
         ];
 
         $issueNumber = null;
-        $exitingIssues = $this->searchApi->issues(sprintf('repo:%s "%s" is:open author:%s', $repository->getFullName(), $title, $this->botUsername));
-        foreach ($exitingIssues['items'] ?? [] as $issue) {
+        $existingIssues = $this->searchApi->issues(sprintf('repo:%s "%s" is:open author:%s', $repository->getFullName(), $title, $this->botUsername));
+        foreach ($existingIssues['items'] ?? [] as $issue) {
             $issueNumber = $issue['number'];
         }
 
@@ -42,6 +42,14 @@ class GithubIssueApi implements IssueApi
             unset($params['labels']);
             $this->issueApi->update($repository->getVendor(), $repository->getName(), $issueNumber, $params);
         }
+    }
+
+    public function lastCommentWasMadeByBot(Repository $repository, $number): bool
+    {
+        $allComments = $this->issueCommentApi->all($repository->getVendor(), $repository->getName(), $number);
+        $lastComment = $allComments[count($allComments) - 1] ?? [];
+
+        return $this->botUsername === ($lastComment['user']['login'] ?? null);
     }
 
     public function show(Repository $repository, $issueNumber): array
@@ -65,5 +73,12 @@ class GithubIssueApi implements IssueApi
             $issueNumber,
             ['body' => $commentBody]
         );
+    }
+
+    public function findStaleIssues(Repository $repository, \DateTimeImmutable $noUpdateAfter): array
+    {
+        $issues = $this->searchApi->issues(sprintf('repo:%s is:issue -linked:pr -label:"Keep open" is:open updated:<%s', $repository->getFullName(), $noUpdateAfter->format('Y-m-d')));
+
+        return $issues['items'] ?? [];
     }
 }
--- a/src/Command/PingStaleIssuesCommand.php
+++ b/src/Command/PingStaleIssuesCommand.php
@@ -1,0 +1,113 @@
+<?php
+
+namespace App\Command;
+
+use App\Api\Issue\IssueApi;
+use App\Api\Issue\IssueType;
+use App\Api\Label\LabelApi;
+use App\Entity\Task;
+use App\Service\RepositoryProvider;
+use App\Service\StaleIssueCommentGenerator;
+use App\Service\TaskScheduler;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * Close issues not been updated in a long while.
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class PingStaleIssuesCommand extends Command
+{
+    public const STALE_IF_NOT_UPDATED_SINCE = '-12months';
+    public const MESSAGE_TWO_AFTER = '+2weeks';
+    public const MESSAGE_THREE_AND_CLOSE_AFTER = '+2weeks';
+
+    protected static $defaultName = 'app:issue:ping-stale';
+
+    private $repositoryProvider;
+    private $issueApi;
+    private $scheduler;
+    private $commentGenerator;
+    private $labelApi;
+
+    public function __construct(RepositoryProvider $repositoryProvider, IssueApi $issueApi, TaskScheduler $scheduler, StaleIssueCommentGenerator $commentGenerator, LabelApi $labelApi)
+    {
+        parent::__construct();
+        $this->repositoryProvider = $repositoryProvider;
+        $this->issueApi = $issueApi;
+        $this->scheduler = $scheduler;
+        $this->commentGenerator = $commentGenerator;
+        $this->labelApi = $labelApi;
+    }
+
+    protected function configure()
+    {
+        $this->addArgument('repository', InputArgument::REQUIRED, 'The full name to the repository, eg symfony/symfony-docs');
+        $this->addOption('dry-run', null, InputOption::VALUE_NONE, 'Do a test search without making any comments or changes');
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        /** @var string $repositoryName */
+        $repositoryName = $input->getArgument('repository');
+        $repository = $this->repositoryProvider->getRepository($repositoryName);
+        if (null === $repository) {
+            $output->writeln('Repository not configured');
+
+            return 1;
+        }
+
+        $notUpdatedAfter = new \DateTimeImmutable(self::STALE_IF_NOT_UPDATED_SINCE);
+        $issues = $this->issueApi->findStaleIssues($repository, $notUpdatedAfter);
+
+        if ($input->getOption('dry-run')) {
+            foreach ($issues as $issue) {
+                $output->writeln(sprintf('Marking issue #%s as "Staled". Link https://github.com/%s/issues/%s', $issue['number'], $repository->getFullName(), $issue['number']));
+            }
+
+            return 0;
+        }
+
+        foreach ($issues as $issue) {
+            $comment = $this->commentGenerator->getComment($this->extractType($issue));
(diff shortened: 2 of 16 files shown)


Nyholm:
This cannot run on a github action like #103 because it needs to create a scheduled task.

wouterj:
I usually don't like auto-closing staled issues: Contrary to staled PRs, there is nothing I as issue owner can do to "unstall" this issue. E.g. I don't know every open source library enough to be able to fix issues I find, so I take the time to open an issue instead. If it's not considered important enough to fix, a maintainer should close it (with a friendly message). If it wasn't closed, it means there is some relevancy of the issue (it just doesn't have enough priority). If it's closed after a year, the issue is there still, it's just hidden.

E.g. some real case: https://github.com/symfony/symfony-docs/issues/4258 This feature is indeed undocumented and useful enough to document. We just haven't yet found the correct example or place in the documentation to fix this issue. That doesn't, imho, mean we should close it.

Another real case I discovered yesterday is for instance https://github.com/SpiderLabs/ModSecurity-nginx/pull/198#issuecomment-644457711 , where a bot autoclosed this issue 2 times already. The PR is just blocked by another PR, there is no reason to close it and "hide" this missing feature.

Nyholm:
It is indeed tricky. But if an issue has not gotten any comments or activity in over a year, it is not likely it will get fixed soon anyways. You can just make a comment to say whatever and the issue will be open for another [time period]. 

I agree with your use case. You as a maintainer can set "Keep open" label on this issue. 

> Another real case I discovered yesterday

I agree with this too. That is why we don't auto close PRs. 

-----

Currently I added a time period for 3 months. Which means, that if there are no activity for 3 months, then we add this comment. We could change this time period to any other value if we think it is too long or too short.

ogizanagi:
I wonder if we really have any benefits re-developing this here, in Carsonbot? 
We mentioned https://github.com/actions/stale some weeks ago about this. Wouldn't it fit?

Nyholm:
Yeah, this is the same as stalebot. But the logic to "redevelop" is way less than 50 lines of code. And we already have a bot =)

Nyholm:
This PR is now "done" and ready for review

jderusse:
I personally don't like auto-closed PR, In my experience, when working with k8s repo, the issues are full of comments "issue stale", "re-open"... It's VERY frustrating for the user that open the PR and for the user that had the exact same issue.

Instead of adding a label to by-pass the bot and keeping the PR open, what's about closing the PR when we don't want to fix it?.

wouterj:
Regarding @jderusse's description: what about changing it from opt-out to opt-in? E.g. have a label that effectively says "hmm, not sure about this one", which will result in auto close within 2 weeks if there is no response.

In the docs, I'm using the `Waiting feedback` for this process manually already.

jderusse:
yeah, a `waiting feedback` on "not reproductible" issues would make more sense to me

Nyholm:
Thank you for the reviews. Note that this is not about PRs. This is about closing issues only. 

Here is an example timeline: 

Day 1: Issue open. 
Day 366: There is no acitivty, Bot makes a comment
Day 370: User makes a comment 
Day 736: There is no activity,  Bot makes another comment
Day 750: Bot closes the issue

I don't think there is a lot of noise... I also don't see no point in having an issue open that there is no activity around... Sure, there are exceptions, that is why we have the "Keep open" label.

Nyholm:
After some discussions with @wouterj, we agree that it is important to consider what we say to the authors. We know they spent time creating this issue. 

After x days of delay: Message 1, should encourage an answer. Maybe the users will close the issue themselves?
After y days of delay: Message 2, comment and say that we are closing the issue. 

Should we make this even more fancy?

After x days of delay: Message 1,
After y days of delay: Message 2, comment and say that we are about to close the issue. 
After z days of delay: Close the issue. 

-------

I need some help improving the content in `StaleIssueCommentGenerator`.

src/Api/Issue/GithubIssueApi.php
```diff
@@ -61,4 +69,11 @@ public function commentOnIssue(Repository $repository, $issueNumber, string $com
             ['body' => $commentBody]
         );
     }
+
+    public function findStaleIssues(Repository $repository, \DateTimeImmutable $noUpdateAfter)
```
wouterj, line 73: Shouldn't this method be added to the `IssueApi` as well?

src/Service/StaleIssueCommentGenerator.php
```diff
@@ -0,0 +1,66 @@
...
+            default:
+                return $this->unknown();
+        }
+    }
+
+    private function bug(): string
+    {
+        return <<<TXT
+Hey,
+Is this bug still relevant? Have you managed to find a workaround?
```
wouterj, line 48:
```suggestion
Hey, thanks for your report!
Is this bug still relevant? Have you managed to find a workaround?
```

(just a suggestion)

src/Service/StaleIssueCommentGenerator.php
```diff
@@ -0,0 +1,66 @@
...
+TXT;
+    }
+
+    private function feature(): string
+    {
+        return <<<TXT
+Thank you for this suggestion.
+Is this still something we want to do?
```
wouterj, line 56: I think the "we want to do" can be understood as "the core team needs to say YES" instead of "I need to engage with this issue". Maybe something like "Would you still like to see this feature?" or the like?

wouterj (commented):
After the discussion with @Nyholm my doubts have somewhat been taken away. The goal here is not to close as many issues as possible, the goal here is to close issues that aren't relevant anymore (e.g. they have been resolved or were no longer needed).

I think the current wording better conveys this message. :+1: I would still like to add something like "There hasn't been much activity here recently" or the like, to indicate the "trigger" for this message.

Nyholm:
Thank you. I've updated the PR accordingly.

wouterj (approved):

src/Api/Issue/GithubIssueApi.php
```diff
@@ -44,6 +44,14 @@ public function open(Repository $repository, string $title, string $body, array
         }
     }
 
+    public function lastCommentWasMadeByBot(Repository $repository, $number): bool
+    {
+        $allComments = $this->issueCommentApi->all($repository->getVendor(), $repository->getName(), $number);
```
Nyholm, line 49: Since this is not using pagination. We will only fetch 100 comments. That means that we will never auto close issues with more that 100 comments. 

Im fine with that atm.

wouterj:
Something I just realized: 300 issues in the code repository and +/- 90 issues in the docs repository are currently within this filter. We should probably organize a small core team sprint day to check these issues before the first deploy closes them.

The advantage of this bot is that it tries to engage the issue creators in determining whether the issue needs to stay open. We would loose that advantage for the biggest set of issues (the first set), if we have to check them all by hand. Maybe we should make the first close window longer (e.g. 4 weeks), in order to give the user the normal 2 weeks to respond and then have 2 weeks for the core team to review the issues were nobody responded? What do you think? After this first round, we can revert the window back to 2 weeks.

Nyholm:
I've added a second message. 

So, Day 1: Make a comment to encourage and action by the user
Day: 15: Do a ping and inform that we will close the issue if we dont hear anything
Day 30: Close the issue.

nicolas-grekas:
What about making this a bit more "manual", eg when an issue has the status "waiting feedback", then we autoclose on 1 month? Applying the label should be manual. I think all issues deserve a comment from someone in the community, no need for the bot to say hello.

Nyholm:
I agree that we could update this in the future. We can also add more rules etc. 
But that is not a blocker from merging this now. The current implementation is the most relaxed and generous rules possible. 

>  I think all issues deserve a comment from someone in the community, no need for the bot to say hello.

If nobody have made a comment in 1 year time, then I think it is fair to say that nobody ever will make a comment.

nicolas-grekas:
> If nobody have made a comment in 1 year time, then I think it is fair to say that nobody ever will make a comment.

Does that happen in practice? Any example?

Nyholm:
Actually yes. Sure it is sad that an issue has been open for long.. 

https://github.com/symfony/symfony/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-asc
- 15014
- 14068
- 19286

wouterj:
More precisely, there are 57 issues without any comment that are open for 1+ year: https://github.com/symfony/symfony/issues?q=is%3Aissue+is%3Aopen+comments%3A0+updated%3A%3C2019-12-09

Nyholm:
> I think all issues deserve a comment from someone in the community, no need for the bot to say hello.

So currently there are 50+ issues that the community has "failed". There is another 250 issues with no activity in the past year. 

These 300 issues will get a "bot comment" to encourage action when this PR is merged.

Nyholm:
I added support for "Stale" label. So now one can easily find all issues that the bot is working with. 

FYI @nicolas-grekas

nicolas-grekas (approved):

src/Command/PingStaleIssuesCommand.php
```diff
@@ -0,0 +1,103 @@
...
+        $this->repositoryProvider = $repositoryProvider;
+        $this->issueApi = $issueApi;
+        $this->scheduler = $scheduler;
+        $this->commentGenerator = $commentGenerator;
+        $this->labelApi = $labelApi;
+    }
+
+    protected function configure()
+    {
+        $this->addArgument('repository', InputArgument::REQUIRED, 'The full name to the repository, eg symfony/symfony-docs');
```
weaverryan, line 48: Sorry, but could we add a `--dry-run` option? And also write output in the command? For example:

> Adding label and comment to issue #555 on symfony/symfony

And, of course, the `--dry-run` would make no changes. It would make me a bit more comfortable, as we could deploy, then run the command manually with `--dry-run` and make sure there are no surprises (like WAY too many issues being commented on, due to some bug).

Thanks!

weaverryan (commented):
Hi! Only a minor comment... mostly to help me deploy more safely. I'd like to deploy this some time during European time, while people are online. And after trying the --dry-run once, probably running the stale-issue command once manually so that the "big dump" of comments can happen in sync, so we can watch it (just in cases).

Thanks!

weaverryan:
Oh, also, could you update the PR title and description to reflect the current way it works?

nicolas-grekas:
> stale-issue command once manually so that the "big dump" of comments can happen in sync, so we can watch it (just in cases).

should we limit the number of notices the bot is going to generate at once?
eg 10 per batch max, one batch per day? (since I suppose this is going to run as a periodic task?)

src/Command/PingStaleIssuesCommand.php
Nyholm: The ´--dry-run´ option is added. 

It will show you 100 issues on Symfony/symfony on the first run. We do not use pagination because it isn't really needed. 

Here is a link to the search: https://github.com/symfony/symfony/issues?q=is%3Aissue+-linked%3Apr+-label%3A%22Keep+open%22+is%3Aopen+updated%3A%3C2019-12-18

Nyholm:
Currently, there is a limit on 100 per batch. We run the script once per day.

Nyholm:
Wohoo. Thank you for merging

PR #110, opened 2020-11-08
Suggest reviewers action

Nyholm:
This PR brings two features: 
1. Someone can make a comment in a PR to say `@carsonbot find me a reviewer please` and a minute later it will reply with a ping to someone.
2. If a PR is opened an had no comments/reviews in 24(?) hours, then we write a comment with a ping to someone. 

This PR works with a Github action that clones the [git-reviewer script](https://github.com/Nyholm/git-reviewer), checks out the source (symfony/symfony or symfony/symfony-docs) and then runs the `git-reviewer`. That will produce a group of users that also have modified the changed files in the PR. 

I ignore FrameworkBundle, all tests and the Changelogs because they are always changed. I also only looking for contributors for the past 2 years. 

When the `git-reviewer` is done, we run the `bin/console app:review:suggest` command. It looks at the contributor lists, removes people (core team) from a block list and removes all people that already have interacted with the PR. 

Then it just writes a comment.


.github/workflows/find-reviewer.yml
```diff
@@ -0,0 +1,88 @@
...
+      - name: Checkout branch base
+        run: |
+          cd ${{ steps.target-repo.outputs.dir }}
+          echo ${{ steps.target-base.outputs.branch }}
+          git pull
+          git checkout ${{ steps.target-base.outputs.branch }}
+
+      - name: Find reviwers
+        env:
+          GITHUB_TOKEN: ${{ secrets.CARSONPROD_GITHUB_TOKEN }}
```
Nyholm, line 72: This needs to be added before we merge.

composer.json
```diff
@@ -2,10 +2,13 @@
     "name": "carsonbot/carsonbot",
     "type": "project",
     "license": "MIT",
+    "repositories": [
+        { "type": "vcs", "url": "https://github.com/Nyholm/php-github-api"}
```
Nyholm, line 6: I hope https://github.com/KnpLabs/php-github-api/pull/931 will be merged and released soon.

config/services.yaml
```diff
@@ -41,6 +41,8 @@ services:
     _defaults:
         autowire: true
         autoconfigure: true
+        bind:
+            string $botUsername: 'nyholm'
```
Nyholm, line 45: This is debug code. Other PRs have added a proper value here.

src/Api/Issue/StdErrIssueApi.php
```diff
@@ -0,0 +1,19 @@
...
+namespace App\Api\Issue;
+
+use App\Model\Repository;
+
+/**
+ * This will not write to Github, only output to STDERR. But it will try to read from Github.
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class StdErrIssueApi extends GithubIssueApi
```
Nyholm, line 12: This class is never used. But it was really useful when testing/debugging, so I kept it in the PR.

composer.json
acrobat: @Nyholm merged and I've release v2.16.0!

Nyholm: Awesome. 
Thank you for the quick merge and release.

Nyholm:
This PR is ready for a review

src/Service/ReviewerFilter.php
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    public function suggestReviewer(array $possibleReviewers, Repository $repository, $pullRequestNumber): ?string
+    {
+        // Dont suggest block listed
+        $possibleReviewers = $this->filterBlocked($possibleReviewers);
+        $possibleReviewers = $this->filterUsersInvolved($possibleReviewers, $repository, $pullRequestNumber);
+
+        foreach ($possibleReviewers as $reviewer) {
+            return $reviewer['username'];
```
Nyholm, line 38: Make sure it has a username

Nyholm: They are already filtered out in `filterBlocked`

javiereguiluz:
Tobias, thanks for proposing this feature. I know that some communities have something similar. MY only fear it's whether this will be truly useful to add reviewers ... or it will be just annoying and won't improve the current situation. But, the only way to know it is to try it for real, so I'm 👍  to try this.

About the proposed command -> `@carsonbot find me a reviewer please`  These "textual interfaces" are always complex to use ... so maybe we could design it with very simple commands/orders instead of textually correct phrases. E.g. `@carsonbot help`, `@carsonbot find reviewer`

Nyholm:
> MY only fear it's whether this will be truly useful to add reviewers

Yeah.. I am not sure.. We will see =)

> About the proposed command 

I agree. The only hard requirements at the moment is to write `@carsonbot` and `review` on the same line. So most phrases will be okey. 

- `@carsonbot review`
- `@carsonbot find reviewer please`
- `@carsonbot get a review`
- `review @carsonbot`
- etc

src/Subscriber/FindReviewerSubscriber.php
```diff
@@ -0,0 +1,86 @@
...
+
+            $number = $data['issue']['number'];
+            $this->pullRequestApi->findReviewer($event->getRepository(), $number, SuggestReviewerCommand::TYPE_DEMAND);
+            $event->setResponseData([
+                'issue' => $number,
+                'suggest-review' => true,
+            ]);
+
+            return;
+        }
```
Nyholm, line 76: FYI @javiereguiluz: Here is the "text command" parsing.

javiereguiluz: Very nice! It follows the *"be liberal in what you accept"* principle.

wouterj: Should we do `preg_match('/\breview\b/', ...)` to prevent `@carsonbot reviewed` from being matched? (e.g. in case someone doesn't understand how to change the status label)

Nyholm: Done. Im using `preg_match` instead. I added a bunch more tests. The same rules apply. If one is using `@carsonbot` and (`review` or `reviewer`) on the same line, then we will trigger a search for a candidate.

.github/workflows/find-reviewer.yml
```diff
@@ -0,0 +1,88 @@
...
+        run: |
+          cd build/reviewer
+          ./git-reviewer.php find ${{ github.event.client_payload.pull_request_number }} ${{ steps.target-repo.outputs.dir }} \
+          --after `date +%Y-%m-%d --date="2 year ago"` \
+          --ignore-path "src/Symfony/FrameworkBundle/*" \
+          --ignore-path "src/Symfony/Bundle/FrameworkBundle/*" \
+          --ignore-path "src/**/Tests/*" \
+          --ignore-path CHANGELOG*.md \
```
stof, line 80: we might also want to ignore the `UPGRADE` file

src/Command/SuggestReviewerCommand.php
```diff
@@ -0,0 +1,110 @@
...
+        $this->repositoryProvider = $repositoryProvider;
+        $this->reviewerFilter = $reviewerFilter;
+        $this->complementGenerator = $complementGenerator;
+    }
+
+    protected function configure()
+    {
+        $this->addArgument('repository', InputArgument::REQUIRED, 'The full name to the repository, eg symfony/symfony-docs');
+        $this->addArgument('number', InputArgument::REQUIRED, 'Pull request number');
+        $this->addArgument('type', InputArgument::REQUIRED, 'Typ is ether "suggest" or "demand".');
```
stof, line 42: typos. `Type` vs `Typ` and `ether` vs `either`

src/Command/SuggestReviewerCommand.php
```diff
@@ -0,0 +1,110 @@
...
+        $contributors = json_decode($json, true);
+        $reviewer = $this->reviewerFilter->suggestReviewer($contributors, $repository, $pullRequestNumber);
+        if (null === $reviewer) {
+            if (self::TYPE_SUGGEST === $type) {
+                $output->writeln('We could not find any reviewer.');
+
+                return 0;
+            }
+
+            $this->issueApi->commentOnIssue($repository, $pullRequestNumber, 'Im sorry. I could not find any suitable reviewer.');
```
stof, line 83: typo `I'm sorry`

Nyholm:
Thank you @stof for the review. I've updated the PR accordingly.

.github/workflows/find-reviewer.yml
```diff
@@ -0,0 +1,89 @@
+name: Find Reviewer
+
+on:
+  repository_dispatch:
+    types: [find-reviewer]
```
weaverryan, line 5: Does this mean this workflow is called each time a `find-reviewer` webhooks is received? Can you help me understand how this works? thx :)

src/Subscriber/FindReviewerSubscriber.php
```diff
@@ -0,0 +1,75 @@
...
+    {
+        $data = $event->getData();
+        if (!in_array($data['action'], ['opened', 'ready_for_review']) || ($data['pull_request']['draft'] ?? false)) {
+            return;
+        }
+
+        $repository = $event->getRepository();
+
+        // set scheduled task to run in 20 hours
+        $this->scheduler->runLater($repository, $data['number'], Task::ACTION_SUGGEST_REVIEWER, new \DateTimeImmutable('+20hours'));
```
weaverryan, line 39: Ok, so when a new PR is opened, or "ready_for_review", then, 20 hours later, we will run the "suggest reviewer" functionality. 20 hours seems fairly fast to me maybe ?

.github/workflows/find-reviewer.yml
ogizanagi: Exact. Here are some resources about it you could find useful:
- https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#repository_dispatch
- https://docs.github.com/en/free-pro-team@latest/rest/reference/repos#create-a-repository-dispatch-event

src/Subscriber/FindReviewerSubscriber.php
ogizanagi: I also wonder what to do with PRs opened with the "Status: Needs Work" label for whatever reasons (or changed with Carson bot). But that's probably the same as opening a Draft PR, we don't want it?

.github/workflows/find-reviewer.yml
stof: see https://github.com/carsonbot/carsonbot/pull/113/files#diff-761989cba708b0fb2f270ad4d799a7a955b1c611d9cefa6fde84611e03e516c0R36

The corresponding code was added in a previous PR, which is indeed weird.

src/Subscriber/FindReviewerSubscriber.php
stof: the status is set to `needs work` when a reviewer asks for changes. Saying we don't want such PRs would be an issue. It means we only accept PRs when they are perfect at first review or when the contributor can change the PR and re-request a review in less than 20 hours after receiving a review (which is putting a lot of burden on them when they contribute in their free time)

.github/workflows/find-reviewer.yml
Nyholm: So I've defined a special event id called `find-reviewer`. In the `GithubPullRequestApi::findReviewer()`  that @stof linked, I am dispatching this special event. 

It is basically me saying: "Hey start the specific CI job called `find-reviewer`".

src/Subscriber/FindReviewerSubscriber.php
Nyholm: @weaverryan Maybe it is too fast, maybe not. Whenever I make a PR I am secretly and eagerly waiting for some kind of feedback. So 20 hours is suuuper long. 

@ogizanagi, I've not been thinking of that.. I should make sure the PR has the correct status. 

A user may open a PR, then change the status them selves. If so, we should not call for a review.

wouterj: I think everyone likes to get a review/comment within a couple hours after opening. However, there is a difference between what we think is reasonable and what we like. 

If I think "when does someone start to add a ping or asks on Slack if someone can please take a look at their PR", I think this margin is closer to 5 days - 1 week than to 20 hours.

Nyholm: Should we try with 20 hours for the first 10 pings? Just to see what happens. 

Also note that none of us never will be pinged. We will mostly ping symfony contributors that nobody ever pinged directly on github or slack.

src/Service/TaskHandler/SuggestReviewerHandler.php
```diff
@@ -0,0 +1,59 @@
...
+            return;
+        }
+
+        $status = $this->statusApi->getIssueStatus($task->getNumber(), $repository);
+        if (Status::NEEDS_REVIEW === $status || null === $status) {
+            $this->pullRequestApi->findReviewer($repository, $task->getNumber(), SuggestReviewerCommand::TYPE_SUGGEST);
+        } else {
+            // Try again later
+            $this->scheduler->runLater($repository, $task->getNumber(), Task::ACTION_SUGGEST_REVIEWER, new \DateTimeImmutable('+20hours'));
+        }
```
Nyholm, line 52: @ogizanagi I've added this logic. 

If the PR does not have status "needs review" we will try again 20 hours later. Just until somebody else has made a comment.

src/Subscriber/FindReviewerSubscriber.php
Nyholm: FYI: The rust community have a bot that pings potential reviewers as soon as the PR is open. 

```
https://github.com/rust-lang/rust/pull/78714
```

src/Service/TaskHandler/SuggestReviewerHandler.php
ogizanagi: 👍  LGTM

src/Service/ReviewerFilter.php
```diff
@@ -0,0 +1,86 @@
...
+ */
+class ReviewerFilter
+{
+    private $issueApi;
+
+    // These users will never be pinged for a review
+    private $blocked = [
+            'fabpot', 'tobion', 'nicolas-grekas', 'stof', 'dunglas', 'jakzal',
+            'xabbuh', 'javiereguiluz', 'lyrixx', 'weaverryan', 'chalasr', 'ogizanagi',
+            'sroze', 'yceruto', 'nyholm', 'wouterj', 'derrabus', 'jderusse',
```
nicolas-grekas, line 23: does the bot have access to the mergers team?

nicolas-grekas:
Does any project use this logic successfully already?
I'm curious about how relevant this will be.
Also: which problem does it solve? Will this solve something that is actually slowing us down right now? That's a honest question :) I feel like things are pretty smooth, but I'm curious to see who this proposal can improve things.
What's the way to verify this provides value? Enable, then wait and see?

src/Service/ReviewerFilter.php
Nyholm: Im not sure. 

But that would be a great future update to implement.

Nyholm:
> Does any project use this logic successfully already?

Yes, @javiereguiluz showed some examples from the Ruby community. They have bot that suggest a reviewer when the PR is opened. 

> which problem does it solve?

Just two very tiny problems. 
1. If a PR dont get any attention, it will try to ping someone
2. It will ping a person that probably never have been pinged before

> What's the way to verify this provides value? Enable, then wait and see?

Yes, Wait and see. Maybe this is a good idea, maybe it is something we disable before Christmas. Let's try something new.

nicolas-grekas (approved):
Let's give it a try :)

weaverryan:
Let's see how it works! Thanks very much Tobias!

PR #115, opened 2020-11-09
Better trim and add a test

Nyholm:
My second attempt to fix the issue in #97. 

The fix in #112 was only fixing part of the problem.

--------

This PR fix the issue when a label is `[Component]&nbsp;Foo`. We trying to change that to `[Component]&nbsp;&nbsp;Foo` (double space). 
Github is nice enough to remove double spaces, but that means that we update the title all the time...


Nyholm:
Thank you for the merge

PR #116, opened 2020-11-10
Fix: Colors are case sensitive...

Nyholm:
Im slowly going to fix all the bugs, here is a start. 

FYI @javiereguiluz, I have the feeling that you add the new component labels. I made it a rule that labels for components must always be grey.


javiereguiluz:
You are right 😅   The color of component labels is `#DDDDDD`

Nyholm:
Yeah. But the few of them I first looked at had color `#dddddd`. ¯\_(ツ)_/¯

weaverryan:
Thank you Tobias!

PR #120, opened 2020-11-12
Make sure we don't update title twice

Nyholm:
I want to avoid this: 

Screenshot 2020-11-12 at 22 38 44

----------
--------
--------
--------

It happens when a user adds two labels at the same time. 

FYI @jderusse

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,6 @@
 {
     "name": "carsonbot/carsonbot",
+    "description": "The Carson Issue Butler",
     "type": "project",
     "license": "MIT",
     "require": {
@@ -70,5 +71,6 @@
             "assets:install %PUBLIC_DIR%": "symfony-cmd"
         }
     },
-    "minimum-stability": "RC"
+    "minimum-stability": "RC",
+    "prefer-stable": true
 }
--- a/composer.json
+++ b/composer.json
@@ -16,8 +16,9 @@
         "symfony/console": "^5.1",
         "symfony/dotenv": "^5.1",
         "symfony/flex": "^1.8",
-        "symfony/framework-bundle": "^5.1",
-        "symfony/http-client": "^v5.2.0-BETA3",
+        "symfony/framework-bundle": "^5.2@rc",
+        "symfony/http-client": "^5.2@rc",
+        "symfony/lock": "^5.2@rc",
         "symfony/monolog-bundle": "~3.5",
         "symfony/security-core": "5.1.*",
         "symfony/twig-pack": "^1.0",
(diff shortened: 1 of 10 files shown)


src/Subscriber/AutoUpdateTitleWithLabelSubscriber.php
```diff
@@ -69,10 +72,22 @@ public function onPullRequest(GitHubEvent $event)
...
+        $lock = $this->lockFactory->createLock($repository->getFullName().'#'.$number);
+        $lock->acquire(true); // blocking
+        try {
+            // Refetch the current title just to make sure it has not changed
+            if ($prTitle === ($this->pullRequestApi->show($repository, $number)['title'] ?? '')) {
+                return;
+            }
+            $this->pullRequestApi->updateTitle($repository, $number, $prTitle);
+        } finally {
+            $lock->release();
```
jderusse, line 86: not needed, Symfony do it for you on `__destruct`

config/packages/lock.yaml
```diff
@@ -0,0 +1,14 @@
+services:
+
+    app.lock.store:
+        class: Symfony\Component\Lock\Store\PdoStore
+        factory: [Symfony\Component\Lock\Store\StoreFactory,  createStore]
+        arguments: ['%env(DATABASE_URL)%']
+
+    app.lock.blocking_store:
```
jderusse, line 8: not needed with symfony/lock 5.2

src/Subscriber/AutoUpdateTitleWithLabelSubscriber.php
Nyholm: Oh, Thank you

config/packages/lock.yaml
Nyholm: I'll have to update then =)

Thank you

Nyholm:
Thank you for the review. The PR is updated

composer.json
```diff
@@ -68,5 +69,6 @@
             "assets:install --symlink --relative %PUBLIC_DIR%": "symfony-cmd",
             "assets:install %PUBLIC_DIR%": "symfony-cmd"
         }
-    }
+    },
+    "minimum-stability": "RC"
```
jderusse, line 73: I would also adds `prefer-stable: true`

symfony.lock
```diff
@@ -268,6 +268,9 @@
     "symfony/http-kernel": {
         "version": "v5.1.2"
     },
+    "symfony/lock": {
+        "version": "v5.1.8"
```
jderusse, line 272: Shouldn't it be 5.2?

jderusse (approved):
Thank you @Nyholm
A nitpicking comment from my side :)

composer.json
Nyholm: I ran composer update with that command option.

symfony.lock
Nyholm: I am not sure. I did not modify this myself. 

I dont think it matters too much.

wouterj (approved):

PR #122, opened 2020-11-13
Make sure to fetch all the labels from Github

Nyholm:
Wohoo. I found a bug. 

When we asked Github for all the labels on symfony/symfony, we just got the 20 first ones. This PR will fix that.


src/Api/Label/GithubLabelApi.php
```diff
@@ -36,11 +37,17 @@ class GithubLabelApi implements LabelApi
...
+     * @var ResultPager
+     */
+    private $resultPager;
+
+    public function __construct(Labels $labelsApi, ResultPager $resultPager, CacheInterface $cache, LoggerInterface $logger)
     {
         $this->labelsApi = $labelsApi;
         $this->cache = $cache;
         $this->logger = $logger;
+        $this->resultPager = $resultPager;
```
weaverryan, line 50: Could you move this up 2 lines so that its order matches the arguments? And the same for the `private $resultPager`

config/services.yaml
```diff
@@ -75,6 +75,8 @@ services:
 
     Github\Api\Issue\Labels:
         factory: ['@Github\Api\Issue', labels]
+        calls:
+            - ['setPerPage', [100]]
```
weaverryan, line 79: We're telling the `Labels` API to return 100 results but we're *also* using `ResultPager` to fetch all the results? That seems potentially redundant - are we doing it just to be safe?

Nyholm: We currently have 87 labels. If we add 14 more this will be a silent error. So yes, the ResultPager is not really needed at the moment.

src/Api/Label/GithubLabelApi.php
Nyholm: Sure

config/services.yaml
Nyholm: Btw, The ResultPager is just making extra requests for us. 

If we left this as (default) 20, the Result pager would make 5 request.

Nyholm:
Thank you for merging

PR #124, opened 2020-11-14
Run a search to verify that this is the first PR of the user

Nyholm:
This will make sure that we don't send a "welcome message" twice when the user's first PR was not merged.


derrabus (approved):

Nyholm:
Thank you for the review

PR #128, opened 2020-11-16
Use dummy token in test to make sure we run functional

Nyholm:
The calls to github is abstracted away


Nyholm:
So.. This is weird. 

With the following config the string "sqlite:///%kernel.project_dir%/var/data.db" is passed to `Symfony\Component\Lock\Store\PdoStore::__construct()`. That will not work since "%kernel.project_dir%" is not resolved...

```
# .env
DATABASE_URL=sqlite:///%kernel.project_dir%/var/data.db

# config/packages/lock.yaml
framework:
    lock: '%env(DATABASE_URL)%'
```

It might be just me that don't understand when and when no to use `resolve` key. I stay away from in because @nicolas-grekas have told me so repeatedly. 

@jderusse, is this expected?

jderusse:
If the content of the env variable contains parameters, you have to use resolve in order to tell Symfony to replace `%foo%` (inside the env variable) by the content of the `foo` parameter.

Nyholm:
I see. And that is why it worked locally for me when I was testing with Postgres, I didn’t have any parameters there.

Nyholm:
Thank you for merging

PR #134, opened 2020-11-17
Add warning when PR is opened to an unsupported branch

### LINKED ISSUE #126: Warn users when selecting an unmaintained branch
The only common issue in symfony-docs contributions is that contributors submit their PRs against an unmaintained branch. Most of the times this is OK because our internal merge tool can switch the branch easily, but sometimes there are lots of merge conflicts.

Ideally, Carson Bot should detect if an unmaintained branch is used and add a comment telling the user about this and linking to https://symfony.com/releases where they can see the currently maintained branches.

Thanks!

Nyholm:
This will fix #126

TODO: 
- [x] Add tests

### DIFF
--- a/src/Service/SymfonyVersionProvider.php
+++ b/src/Service/SymfonyVersionProvider.php
@@ -26,7 +26,7 @@ class SymfonyVersionProvider
         $this->cache = $cache;
     }
 
-    public function getCurrentVersion()
+    public function getCurrentVersion(): string
     {
         $httpClient = $this->httpClient;
 
@@ -40,9 +40,25 @@ class SymfonyVersionProvider
                 $version = $defaultValue;
             }
 
-            $item->expiresAfter($version === $defaultValue ? 300 : 604800);
+            $item->expiresAfter($version === $defaultValue ? 300 : 86400);
 
             return $version;
         });
+    }
+
+    /**
+     * @throws \RuntimeException
+     */
+    public function getSupportedVersions(): array
+    {
+        $response = $this->httpClient->request('GET', 'https://symfony.com/releases.json');
+        $data = $response->toArray(true);
+        $versions = $data['supported_versions'] ?? null;
+
+        if (is_array($versions) && count($versions) > 0) {
+            return $versions;
+        }
+
+        throw new \RuntimeException('Could not fetch supported versions');
     }
 }
--- a/src/Subscriber/UnsupportedBranchSubscriber.php
+++ b/src/Subscriber/UnsupportedBranchSubscriber.php
@@ -1,0 +1,88 @@
+<?php
+
+namespace App\Subscriber;
+
+use App\Api\Issue\IssueApi;
+use App\Event\GitHubEvent;
+use App\GitHubEvents;
+use App\Service\ComplementGenerator;
+use App\Service\SymfonyVersionProvider;
+use Psr\Log\LoggerInterface;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+/**
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class UnsupportedBranchSubscriber implements EventSubscriberInterface
+{
+    private $symfonyVersionProvider;
+    private $issueApi;
+    private $complementGenerator;
+    private $logger;
+
+    public function __construct(SymfonyVersionProvider $symfonyVersionProvider, IssueApi $issueApi, ComplementGenerator $complementGenerator, LoggerInterface $logger)
+    {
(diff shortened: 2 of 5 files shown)


src/Service/SymfonyVersionProvider.php
```diff
@@ -45,4 +45,32 @@ public function getCurrentVersion()
...
     }
+
+    /**
+     * @throws \RuntimeException
+     */
+    public function getSupportedVersions(): array
+    {
+        $httpClient = $this->httpClient;
+
+        $versions = $this->cache->get('symfony_version', function (ItemInterface $item) use ($httpClient) {
```
noniagriconomie, line 56: isnt this the same key as above?

src/Subscriber/UnsupportedBranchSubscriber.php
```diff
@@ -0,0 +1,79 @@
...
+        $this->complementGenerator = $complementGenerator;
+    }
+
+    /**
+     * Sets milestone on PRs that target non-default branch.
+     */
+    public function onPullRequest(GitHubEvent $event)
+    {
+        $data = $event->getData();
+        $repository = $event->getRepository();
```
noniagriconomie, line 35: can be moved bellow where used, useless in case of return :)

src/Subscriber/UnsupportedBranchSubscriber.php
```diff
@@ -0,0 +1,79 @@
...
+
+        $number = $data['pull_request']['number'];
+        $complement = $this->complementGenerator->getPullRequestComplement();
+        $validBranchesString = implode(', ', $validBranches);
+        $this->issueApi->commentOnIssue($repository, $number, <<<TXT
+Hey!
+
+$complement
+
+But you have made this PR towards a branch that is not maintain anymore. :/
```
noniagriconomie, line 58:
```suggestion
But you have made this PR towards a branch that is not maintained anymore. :/
```

src/Subscriber/UnsupportedBranchSubscriber.php
```diff
@@ -0,0 +1,79 @@
...
+Could you update the PR base to target one of these branches: $validBranchesString.
```
noniagriconomie, line 59:
```suggestion
Could you update the [PR base branch](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/changing-the-base-branch-of-a-pull-request) to target one of these branches instead? $validBranchesString.
```

noniagriconomie (commented):
small review checking this project :)

src/Service/SymfonyVersionProvider.php
Nyholm: Ooops, you are correct.

src/Subscriber/UnsupportedBranchSubscriber.php
Nyholm: Thank you

Nyholm (commented):
Thank you for the review!

src/Service/SymfonyVersionProvider.php
```diff
@@ -45,4 +45,32 @@ public function getCurrentVersion()
...
+        $versions = $this->cache->get('symfony_supported_versions', function (ItemInterface $item) use ($httpClient) {
+            try {
+                $response = $httpClient->request('GET', 'https://symfony.com/releases.json');
+                $data = $response->toArray(true);
+                $versions = $data['supported_versions'] ?? null;
+            } catch (\Throwable $e) {
+                $versions = null;
+            }
+
+            $item->expiresAfter(null === $versions ? 300 : 604800);
```
weaverryan, line 65: What would happen the first 7 days after, for example, 5.3 is released? Actually, I don't know the exact moment that a new release is added to `releases.json`, so I'm wondering if - especially with the caching - if there will be some time when a version is supported and making PR to it IS valid... but it's not in the version yet. Also... I think the new branches (e.g. 5.3) is created a few weeks before the stable release.

nicolas-grekas:
Yes. Actually, this should close immediately instead of adding a warning.

src/Service/SymfonyVersionProvider.php
Nyholm: We could relax this a bit more. Ie, shorter cache. 

Two things can happen: 
1) A user submits a PR to a branch that is not supported and we dont make a comment.
2) A user submits a PR to a branch that is just created (5.3) and we do make a comment that it is not supported. 

Number 2 is an issue, but if that happens the first day.. then it is fine. I'll reduce this to 1 day

weaverryan:
Thanks Tobias!

PR #138, opened 2020-11-20
Update log configuration to match SymfonyCloud recos

tucksaun:
See https://symfony.com/doc/current/cloud/cookbooks/go_live.html#logging


Nyholm (approved):
Thank you

PR #141, opened 2020-12-05
Open issues for missing config reference

Nyholm:
Should I use this for other bundles too?


.github/workflows/docs-configuration-reference.yml
```diff
@@ -0,0 +1,73 @@
...
+
+      - name: Checkout Symfony Docs repo
+        run: git clone https://github.com/symfony/symfony-docs .github/workflows/docs-configuration-reference/docs
+
+      - name: Download dependencies
+        run: composer install --no-interaction --optimize-autoloader
+
+      - name: Verify docs
+        env:
+          GITHUB_TOKEN: ${{ secrets.CARSONPROD_GITHUB_TOKEN }}
```
Nyholm, line 49: This secret needs to be added

Nyholm: Maybe it has been added already?

weaverryan: Yes, this does exist already on this repository

weaverryan (commented):
Is this still blocked? Do we need to upgrade dependencies?

Also, it strikes me that this may be more useful as a dashboard that is up-to-date and can be checked than something where we create issues each time we find something missing - basically like your translations dashboard.

.github/workflows/docs-configuration-reference.yml
Nyholm: Thank you

Nyholm:
> Is this still blocked? Do we need to upgrade dependencies?

Nope, it is not blocked any more. 

We dont need to bump dependencies. The CI job will checkout master of Symfony/symfony. 

> Also, it strikes me that this may be more useful as a dashboard that is up-to-date and can be checked than something where we create issues each time we find something missing - basically like your translations dashboard.

I agree. Im also happy to make the translation dashboard part of Carson. The issue with having only a dashboard, is that we get multiple issue trackers. 

Anyhow, dashboard or not, I think this PR is a good first step.

OskarStark:
Very nice @Nyholm could you add a new test-repo, run the command and create the potential newly created issues there? In this case we can double check if everything went well and don't bother all the symfony-docs users ?

Nyholm:
Sure. Here is an issue generated by this script. It is for Framework bundle
https://github.com/carsonbot-playground/symfony/issues/13

Nyholm:
I see that the XML dumper is not including "Examples" or "Prototype". 

That will make our output very different. See the Security bundle. 
https://github.com/carsonbot-playground/symfony/issues/19
https://github.com/carsonbot-playground/symfony/issues/16

OskarStark:
Is this still needed ?

Nyholm:
It is a nice feature to tell us where we have missing documentation. .

PR #142, opened 2020-12-05
Make sure we consider case when updating title

Nyholm:
Just a small bugfix. To avoid names like: 

```
[TwigBundle] [twigbundle] Fixed syntax error in config
```

#SymfonyHackday


Nyholm:
Thank you for merging

PR #155, opened 2021-01-07
Remove complements

Nyholm:
Update: 

This PR will remove all bot complements


src/Service/ComplementGenerator.php
```diff
@@ -34,6 +34,11 @@ public function getPullRequestComplement()
             'Excellent, keep up the good work.',
             'Two days ago, I was sitting at my usual spot at the top of Big Ben, I was thinking that we really needed this. And now, here you are with a PR. =)',
             'Great work. I told my friends about this PR, they too were impressed.',
+            'Fantastic! I could not have done it better myself.',
+            'Wow, I like what you are doing here.',
+            'When I first reviewed this, I didn\'t really understand it. But now I do. Thank you.',
+            'It looks like you have taken everything they teach at programming school and put it into practice. More of this please =)',
+            'I have reviewed this PR for over 2M clock cycles now and I am really happy with what I see.',
```
weaverryan, line 41: These are really fun, but the compliments in general seem out-of-context. They are used in 3 spots:

1) When suggesting a reviewer
2) When closing a draft PR
3) When a PR is made to an unsupported branch

The last 2 do not seem right for these compliments - as we're ultimately closing. I think it would be ok to say "thanks for opening the PR", but something like "This is amazing! Thanks!" followed by "now we will close this" feels dishonest/robotic.

For the 1 case of suggesting a reviewer, I also think many of these aren't quite right. Saying "I'm impressed by this PR" from a bot is disingenuous. I think having fun with the idea of carson being a "bot" is ok - this last one (about 2M clock cycles) is probably the best.

WDYT?

Nyholm: > The last 2 do not seem right for these compliments - as we're ultimately closing.

Im not sure I agree. Number 2 is not about "closing a draft PR". It is about encourage to click on "ready for review". A complement is good in that case. 

Number 3 is: 

```
Hey!
$complement
But you have made this PR towards a branch that is not maintained anymore. :/
```

---------

> Saying "I'm impressed by this PR" from a bot is disingenuous.

Could you elaborate? Is it because no bot can be impressed? 

I like these messages to be a bit different every time because it feel less like a bot and they are somewhat fun. Im trying to give Carson some kind of personality and that means he is changing his messages. 



------

FYI, At Happyr, we got a similar bot called Leeroy. It can fix code style if you ask it to in a PR comment. However, sometimes, he will decline your request if you don't say "please".

fabpot: I agree that the bot should not give their feeling about a PR as it does not understand the code. Having a bot say "Great work" and a few hours later a human close the PR because it is wrong is a bit weird.

Nyholm:
I've updated the PR. It will now remove all complements.

chalasr (approved):
Thank you Tobias.

weaverryan:
Thanks Tobias!

PR #159, opened 2021-01-16
Reduce the time before the first ping

Nyholm:
We've tried 12 months. From what I've seen there have only been positive responses. A bunch of issues have been closed but nobody was upset. 

People also understood that they should make a comment to show that the issue is still relevant and maintainer have added the "Keep open" label on some issues. 

Even I have been notified about issues I've forgotten. =)

The original request for this feature suggested 3 months before the first ping, we decided to try out this feature on super old issues first. I think the try out period has proven to be successful.

### DIFF
--- a/src/Command/PingStaleIssuesCommand.php
+++ b/src/Command/PingStaleIssuesCommand.php
@@ -61,7 +61,9 @@ class PingStaleIssuesCommand extends Command
             return 1;
         }
 
-        $notUpdatedAfter = new \DateTimeImmutable((string) $input->getOption('not-updated-for'));
+        /** @var string $timeString */
+        $timeString = $input->getOption('not-updated-for');
+        $notUpdatedAfter = new \DateTimeImmutable($timeString);
         $issues = $this->issueApi->findStaleIssues($repository, $notUpdatedAfter);
 
         if ($input->getOption('dry-run')) {
--- a/src/Command/PingStaleIssuesCommand.php
+++ b/src/Command/PingStaleIssuesCommand.php
@@ -63,7 +63,7 @@ class PingStaleIssuesCommand extends Command
 
         /** @var string $timeString */
         $timeString = $input->getOption('not-updated-for');
-        $notUpdatedAfter = new \DateTimeImmutable($timeString);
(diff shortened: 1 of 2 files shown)


chalasr (approved):

wouterj:
First: I do share your observations about this feature. The feature works much better than I honestly expected.

However, I find 3 months to be on the far end of the "time gradient". 1 year maybe is too long (we can get user interaction quicker, but 3 months is even shorter than the development phase of a single release.

That means that if I post a feature request in the first week of a release, I need to have it finished one month before feature freeze in order not to make Carson ask for interaction. Even more, if I respond to Carson, but I do not manage to finish before feature freeze (and the feature is postponed one release), I'll get another question from Carson before development phase of the next release begins.

I think following the release cycle (6 months) or maybe 5 months should be the absolute minimum for it not to be annoying even for me (a regular contributor and a not-so-common issue opener).

---

I think for docs 1 year is exactly the correct length (less resources & less contributors). So I would love to see this being configured per repo.

Nyholm:
> The feature works much better than I honestly expected.

Same. Im happy =)

-------

I think your reasoning makes sense. I've updated the PR accordingly.

src/Command/PingStaleIssuesCommand.php
```diff
@@ -61,7 +61,9 @@ protected function execute(InputInterface $input, OutputInterface $output)
             return 1;
         }
 
-        $notUpdatedAfter = new \DateTimeImmutable(self::STALE_IF_NOT_UPDATED_SINCE);
+        /** @var string $timeString */
+        $timeString = $input->getOption('not-updated-for');
+        $notUpdatedAfter = new \DateTimeImmutable($timeString);
```
wouterj, line 66:
```suggestion
        $notUpdatedAfter = new \DateTimeImmutable('-'.ltrim($timeString, '-'));
```

I believe this has to be negative (given the previous const value), but the option is as a positive value.

Nyholm: That is true. I forgot about that. 

You cannot have an argument or option value starting with a dash.. the console component will correctly intrepid that as a new option. 

I'll update the PR.

wouterj (approved):

PR #165, opened 2021-02-17
Update title when labels are removed

Nyholm:
This will fix #151

PR #167, opened 2021-02-18
Bugfix with auto label

Nyholm:
Fixed issue introduced in #164 and #165

This is a nasty bug, it is all about race condition and it depends on what webhook is sent when. 
When a PR is opened, there is usually 5 or so webhook requests to carson. Some includes all lables, some do not. This PR make sure we query Github for the latest data before we make any changes.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,6 +29,7 @@
         "symfony/symfony": "*"
     },
     "require-dev": {
+        "happyr/service-mocking": "^0.1.3",
         "symfony/browser-kit": "^5.1",
         "symfony/phpunit-bridge": "^5.1",
         "symfony/web-profiler-bundle": "^5.1"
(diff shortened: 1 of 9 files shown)


Nyholm:
Im not sure if the extra sleep is needed, but I do know github have some issues with syncing their APIs when there is a lot happening at the same time.

tests/Subscriber/AutoUpdateTitleWithLabelSubscriberTest.php
```diff
@@ -88,42 +82,14 @@ public function testOnPullRequestLabeledCaseInsensitive()
...
-        $responseData = $event->getResponseData();
-        $this->assertEmpty($responseData);
-    }
-
-    /**
-     * If a user add two labels at the same time. We will get two webhooks simultaneously.
-     * We need to make sure that when we request the Github API it will return the changes
-     * from the first webhook.
-     */
-    public function testOnPullRequestLabeledTwice()
```
Nyholm, line 113: This test was removed because if synced the PR data (for "title") with the webhook data. We are not using the webhook data anymore.

composer.json
```diff
@@ -29,6 +29,7 @@
         "symfony/symfony": "*"
     },
     "require-dev": {
+        "happyr/service-mocking": "^0.1.3",
```
Nyholm, line 32: I like this package more and more

PR #168, opened 2021-03-03
Add comment about squashing commits

Nyholm:
Scenario 1) If a user opens a PR with multiple commits and don't allow maintainers to edit the branch. Then `gh` will squash the commits and **close** the PR. 

Scenario 2) If there are multiple authors of the commits in the PR, the maintainer need to squash the commits manually, push to the fork and then merge the PR. 

With these scenarios in mind, I think it is a good thing to let Carson tell the user that they need to squash the commits themselves or allow the maintainer to do it.

### DIFF
--- a/src/Subscriber/AllowEditFromMaintainerSubscriber.php
+++ b/src/Subscriber/AllowEditFromMaintainerSubscriber.php
@@ -1,0 +1,59 @@
+<?php
+
+namespace App\Subscriber;
+
+use App\Api\Issue\IssueApi;
+use App\Api\PullRequest\PullRequestApi;
+use App\Event\GitHubEvent;
+use App\GitHubEvents;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+/**
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class AllowEditFromMaintainerSubscriber implements EventSubscriberInterface
+{
+    private $commentsApi;
+    private $pullRequestApi;
+
+    public function __construct(IssueApi $commentsApi, PullRequestApi $pullRequestApi)
+    {
+        $this->commentsApi = $commentsApi;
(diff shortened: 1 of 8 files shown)


src/Subscriber/AllowEditFromMaintainerSubscriber.php
```diff
@@ -0,0 +1,59 @@
...
+        $repository = $event->getRepository();
+        $pullRequestNumber = $data['pull_request']['number'];
+        $this->commentsApi->commentOnIssue($repository, $pullRequestNumber, <<<TXT
+Please note that you need squash your commits before this PR can be merged. The maintainer can also squash the commits for you, but then you need to “Allow edits from maintainer” (there is a checkbox in the sidebar of the PR).
+
+Cheers!
+
+Carsonbot
+TXT
+);
```
derrabus, line 45: You can indent HEREDOC blocks in current versions of PHP. 🙂

Nyholm: ¯\_(ツ)_/¯

stof:
Should we put such comment only on PRs having multiple commits in them ?

derrabus:
> Should we put such comment only on PRs having multiple commits in them ?

The contributor might add more commits later. So it might be reasonable emit that message even on single-commit PRs.

chalasr:
Should we just add it to the PR template then?

Nyholm:
Detecting multiple commits is more correct. But that increases complexity by a lot. We need to both check when PR open and when PR got a push. We also need to to use `git` to figure out how many commits there are, Im not sure that data in in the webhook payload. At last we need to keep track of if we made this comment on the PR earlier or not. Bottom line is, I really dont want to implement this feature =)

> Should we just add it to the PR template then?

What should we add? I think the PR template is good as it is. Especially if Carson makes this comment.

PR #169, opened 2021-03-12
Ignore changes to composer.json when looking for reviewer

Nyholm:
See https://github.com/symfony/symfony/pull/40436#issuecomment-796610022

### DIFF
--- a/.github/workflows/find-reviewer.yml
+++ b/.github/workflows/find-reviewer.yml
@@ -84,11 +84,11 @@ jobs:
           cd build/reviewer
           ./git-reviewer.php find ${{ github.event.client_payload.pull_request_number }} ${{ steps.target-repo.outputs.dir }} \
           --after `date +%Y-%m-%d --date="2 year ago"` \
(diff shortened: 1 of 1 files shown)


.github/workflows/find-reviewer.yml
```diff
@@ -84,11 +84,11 @@ jobs:
           cd build/reviewer
           ./git-reviewer.php find ${{ github.event.client_payload.pull_request_number }} ${{ steps.target-repo.outputs.dir }} \
           --after `date +%Y-%m-%d --date="2 year ago"` \
-          --ignore-path "src/Symfony/FrameworkBundle/*" \
```
Nyholm, line 87: This path is not valid.

wouterj:
Wouldn't it be better to say "only include `*.php` files (excluding tests)`?

Nyholm:
Yeah you are correct. But that requires me to update the `git-reviewer` repo. 

This change would be pretty much the same.

Issue #179, opened 2021-05-31
Cannot create task


Nyholm:
This error shows up in the log: 

>  app.ERROR: Failed dispatching "pull_request" event for "carsonbot-playground/symfony" repository. {"exception":"[object] (Doctrine\\DBAL\\Exception\\NotNullConstraintViolationException(code: 0): An exception occurred while executing 'INSERT INTO task (repository_full_name, number, action, created_at, updated_at, verify_after) VALUES (?, ?, ?, ?, ?, ?)' with params [\"carsonbot-playground\\/symfony\", 22, 4, \"2021-05-31 14:10:11\", \"2021-05-31 14:10:11\", \"2021-06-01 10:10:11\"]:\n\nSQLSTATE[23502]: Not null violation: 7 ERROR:  null value in column \"id\" violates not-null constraint\nDETAIL:  Failing row contains (null, carsonbot-playground/symfony, 22, 4, 2021-05-31 14:10:11, 2021-05-31 14:10:11, 2021-06-01 10:10:11). at /app/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/AbstractPostgreSQLDriver.php:64)\n[previous exception] [object] (Doctrine\\DBAL\\Driver\\PDO\\Exception(code: 23502): SQLSTATE[23502]: Not null violation: 7 ERROR:  null value in column \"id\" violates

Nyholm:
PR #181 fixed the issue. But Im not sure why..

Nyholm:
Using doctrine/doctrine-bundle 2.3.2 is bad. The error exists in that version.

Nyholm:
Using doctrine/doctrine-bundle 2.2.4 is okey.

Nyholm:
Using doctrine/doctrine-bundle 2.3.0 is bad. This version is the first version that is broken.

stof:
@Nyholm can you try configuring the db driver and server versions explicitly in the DBAL config instead of extracting them from the credentials URL ? DoctrineBundle 2.3.0 introduces a new metadata caching configuration that automatically cache the mapping during the cache-warmup (enabled only in non-debug mode, which might explain why you don't see the issue in dev). If you run the cache-warmup without the DB credentials in the env variables, the ORM might receive the wrong DB platform when generating the metadata, which would cache the wrong metadata (and so potentially break things)

stof:
[...]

PR #189, opened 2021-06-02
Use doctrine-bundle 2.3 and disable metadata cache

Nyholm:
Related to #179

This should make sure we dont store any metadata about the entities.


Nyholm:
I changed my mind. I was able to reproduce the issue locally. 

```
rm -rf var/cache
# make sure the only database is the sqllite in .env
php bin/console cache:clear --no-warmup --env=prod
# Add a postgress DSN (.env.local)
# Write to database
-- Error
```


This will not work.

PR #190, opened 2021-06-03
Specify production values for doctrine dbal

Nyholm:
Related to #179

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
         "php": ">=7.4",
         "ext-json": "*",
         "composer/package-versions-deprecated": "^1.11",
-        "doctrine/doctrine-bundle": "2.2.4",
+        "doctrine/doctrine-bundle": "2.3.0",
         "doctrine/doctrine-migrations-bundle": "^3.0",
         "doctrine/dbal": "^2.13.1",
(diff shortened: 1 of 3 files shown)


nicolas-grekas (approved):

composer.json
```diff
@@ -7,7 +7,7 @@
         "php": ">=7.4",
         "ext-json": "*",
         "composer/package-versions-deprecated": "^1.11",
-        "doctrine/doctrine-bundle": "2.2.4",
+        "doctrine/doctrine-bundle": "2.3.0",
```
stof, line 10: why keeping an exact version now ? Are you still bisecting ?

Nyholm: I just wanted to confirm that this was a fix. I'll follow this up with a second pr to allow the latest versions of all doctrine packages.

Nyholm:
Interesting. This PR did not fix the issue..

stof:
@Nyholm try putting a postgresql string in .env. I think the .env value will still win over this explicit config as DBAL makes the URL win.

Nyholm:
Yepp. I believe you are correct: https://github.com/symfony-tools/carsonbot/pull/192

PR #191, opened 2021-06-07
Rewrite unwanted phrases

Nyholm:
Let's make sure no PR is merged with phrases we don't appreciate.


src/Subscriber/RewriteUnwantedPhrasesSubscriber.php
```diff
@@ -0,0 +1,80 @@
...
+    {
+        $this->pullRequestApi = $pullRequestApi;
+        $this->lockFactory = $lockFactory;
+    }
+
+    public function onPullRequest(GitHubEvent $event)
+    {
+        $data = $event->getData();
+        $action = $data['action'];
+        if (!in_array($action, ['opened', 'ready_for_review', 'edited'])) {
```
weaverryan, line 32: This type of check needs to be centralized. We've got `ready_for_review`, for example, in 7 different subscribers now :) 

Perhaps we could introduce a "model" class that we pass the `GitHubEvent` into (e.g. `new GithubPullRequestData($event)`) and then we can call things like `$pullRequest->isInOpenState())` or something similar.

Can be done in another PR.

src/Subscriber/RewriteUnwantedPhrasesSubscriber.php
```diff
@@ -0,0 +1,80 @@
...
+            // Only update PullRequests
+            return;
+        }
+
+        $repository = $event->getRepository();
+        $number = $data['number'];
+
+        sleep(1); // Wait for github API to be updated
+        $lock = $this->lockFactory->createLock($repository->getFullName().'#'.$number);
+        $lock->acquire(true); // blocking. Lock will be released at __destruct
```
weaverryan, line 46: This is the first time that we've needed to lock and sleep? What makes this subscriber unique in that way?

Nyholm: The second time: https://github.com/symfony-tools/carsonbot/blob/a08565befe2aed5e2c730523c24f89b805e5e983/src/Subscriber/AutoUpdateTitleWithLabelSubscriber.php#L52

Right when we open a PR we do many requests to modify the PR. Which will sometimes lead to race conditions. 

This new subscriber is very similar to the AutoUpdateTitleWithLabelSubscriber

Nyholm:
I agree that we may introduce some more objects, we may also make the subscribers fail individually and try to minimise "update requests". But that fall outside of the current PR.

src/Subscriber/RewriteUnwantedPhrasesSubscriber.php
weaverryan: Ah, indeed - I had somehow missed this same code in `AutoUpdateTitleWithLabelSubscriber`. That's maybe another thing that could be centralized a bit... later ;)

PR #198, opened 2021-11-06
Add PR message if branch desc mismatch branch target

94noni:
Hi,

Small PR to try https://github.com/symfony-tools/carsonbot/issues/176

I may need info on how I can test this with the repo `carsonbot-playground/symfony`

Also for now the branch extract is very raw, maybe you have some idea to improve :)


composer.json
```diff
@@ -7,10 +7,10 @@
         "php": ">=7.4",
         "ext-json": "*",
         "composer/package-versions-deprecated": "^1.11",
+        "doctrine/common": "^3.0",
```
94noni, line 10: this was done via the `"sort-packages": true`

src/Subscriber/MismatchBranchDescriptionSubscriber.php
```diff
@@ -0,0 +1,85 @@
...
+        $targetBranch = $data['pull_request']['base']['ref'];
+        if ($targetBranch === $descriptionBranch) {
+            return;
+        }
+
+        $this->issueApi->commentOnIssue($event->getRepository(), $number, <<<TXT
+Hey!
+
+Thanks for your PR. You are targeting branch "$targetBranch" but it seems your PR description refers to branch "$descriptionBranch".
+Could you update the PR description please? This helps core maintainers a lot.
```
Nyholm, line 50: Or maybe the target branch is wrong?

Nyholm (commented):
Thank you
We still need to merge and deploy to use the playground.

Could you please write an integration test for this feature? (See and copy an existing one)

PR #205, opened 2022-05-02
Fixed static analysers

Nyholm:
These sort of PRs are usually opinionated. I make this change because our main branch is currently red.

### DIFF
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -53,7 +53,7 @@ jobs:
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: 7.4
+          php-version: 8.1
           coverage: none
           tools: vimeo/psalm:4.22.0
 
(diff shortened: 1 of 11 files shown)


.github/workflows/static.yml
```diff
@@ -9,54 +9,59 @@ on:
...
       - name: Download dependencies
-        run: composer install --no-interaction --optimize-autoloader
+        uses: ramsey/composer-install@v1
 
       - name: PHPStan
-        uses: docker://oskarstark/phpstan-ga:0.12.48
-        with:
-          entrypoint: /composer/vendor/bin/phpstan
-          args: analyze --no-progress
+        run: phpstan analyze --no-progress --error-format=checkstyle | cs2pr
```
stof, line 29: no need to use `--error-format=checkstyle | cs2pr`. phpstan supports github annotations natively since a long time.

Nyholm: Yeah, I know. But there is a slight difference last I checked. I dont remember what it was now.

PR #212, opened 2022-12-05
Updating dependencies and php version

weaverryan:
Currently blocked by https://github.com/Nyholm/git-reviewer/issues/4 as we have one GH action that runs both Carson AND that library at the same time. And so, they both need to support 8.1

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,25 +7,26 @@
         "php": ">=8.1",
         "ext-json": "*",
         "composer/package-versions-deprecated": "^1.11",
+        "doctrine/common": "^3.0",
+        "doctrine/dbal": "^2.13",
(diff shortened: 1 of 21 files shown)


composer.json
```diff
@@ -40,7 +40,7 @@
     },
     "config": {
         "platform": {
-            "php": "7.4"
+            "php": "8.1."
```
OskarStark, line 43:
```suggestion
            "php": "8.1"
```
?

Nyholm:
Let's drop the git-reviewer feature.

Nyholm:
PR is ready for a review.

OskarStark (approved):
Should we switch to 8.2 ?

Nyholm:
Yes. But I let that be the next PR

PR #216, opened 2023-03-20
add labels for new components

xabbuh:


weaverryan:
Thanks @xabbuh! Carson deploys on merge automatically - so this should be up there shortly

xabbuh:
@weaverryan Do you by chance have an idea why labels are not (always) added by Carson (see symfony/symfony#49804 for an example)?

PR #232, opened 2024-08-12
replace docker-compose with docker compose

xabbuh:

PR #233, opened 2024-08-12
fix: docker compose on CI

alexislefebvre:
and use full name for option (`--detach` instead of `-d` that is not straightforward)


xabbuh:
see #232 :)

alexislefebvre:
> see #232 :)

I have been too slow. :laughing:

PR #234, opened 2024-08-18
Trim whitespace better

Nyholm:
We still see whitespace characters in titles. Example: https://github.com/symfony/symfony/pull/57753#event-13537603748

We did an attempt to fix it in #223. 

I suspect we have some HTML characters, so I convert them to strings, then do a more aggressive trim with regex.

### DIFF
--- a/src/Subscriber/AutoUpdateTitleWithLabelSubscriber.php
+++ b/src/Subscriber/AutoUpdateTitleWithLabelSubscriber.php
@@ -76,8 +76,11 @@ class AutoUpdateTitleWithLabelSubscriber implements EventSubscriberInterface
             $prPrefix .= '['.$label.']';
         }
 
+        // Clean string from all HTML chars and remove whitespace at the beginning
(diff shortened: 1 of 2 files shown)


src/Subscriber/AutoUpdateTitleWithLabelSubscriber.php
```diff
@@ -76,8 +76,11 @@ public function onPullRequest(GitHubEvent $event)
             $prPrefix .= '['.$label.']';
         }
 
+        // Clean string from all HTML chars and remove whitespace at the beginning
+        $prTitle = preg_replace('@^[\h\s]+@u', '', html_entity_decode($prTitle));
+
         // Add back labels
```
OskarStark, line 82: Outdated comment?

Nyholm: No, I dont think so. The `$prPrefix` variables has the labels.

PR #235, opened 2024-08-20
Update StaleIssueCommentGenerator.php

94noni:
relates to https://github.com/symfony/symfony/issues/50778
just a raw idea like this
I agree with the [comment](https://github.com/symfony/symfony/issues/50778#issuecomment-2200191716), sometime people comment and would like a feature but perhaps dont know that this framework is community driven first

### DIFF
--- a/src/Service/StaleIssueCommentGenerator.php
+++ b/src/Service/StaleIssueCommentGenerator.php
@@ -71,6 +71,9 @@ TXT;
         return <<<TXT
 Thank you for this suggestion.
 There has not been a lot of activity here for a while. Would you still like to see this feature?
+[Every feature request may or may not be developed in future versions based on the community.](https://symfony.com/doc/current/contributing/community/releases.html#development)
+Perhaps someone would like to try?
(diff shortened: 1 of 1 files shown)


src/Service/StaleIssueCommentGenerator.php
```diff
@@ -71,6 +71,9 @@ private function feature(): string
         return <<<TXT
 Thank you for this suggestion.
 There has not been a lot of activity here for a while. Would you still like to see this feature?
+[Every feature request may or may not be developed in future versions based on the community.](https://symfony.com/doc/current/contributing/community/releases.html#development)
```
Nyholm, line 74: I see you took that sentence from the website. But it is a strange sentence.. Maybe we can make it simpler?

```suggestion
[Every feature is developed by the community.](https://symfony.com/doc/current/contributing/community/releases.html#development)
```

Nyholm (commented):
Thank you. 

I appreciate this change.

src/Service/StaleIssueCommentGenerator.php
94noni: Agreed

Nyholm (approved):
Thank you

PR #240, opened 2024-09-16
Cleanup migrations

OskarStark:


migrations/Version20201112220914.php
```diff
@@ -36,6 +26,6 @@ public function up(Schema $schema): void
 
     public function down(Schema $schema): void
     {
-        // this down() migration is auto-generated, please modify it to your needs
+        $this->throwIrreversibleMigrationException();
```
xabbuh, line 29: What about dropping the table instead?

OskarStark: I don't like down() migrations and I don't think we want to do this anyway

Nyholm (approved):

PR #246, opened 2024-09-17
Execute migrations in CI

OskarStark:


.github/workflows/ci.yml
```diff
@@ -31,7 +31,7 @@ jobs:
         run: composer install --no-interaction --optimize-autoloader
 
       - name: Setup database
-        run: bin/console doctrine:schema:create
+        run: bin/console doctrine:migrations:migrate --no-interaction
```
Nyholm, line 34: Why would I like this? 

It just takes more time, right? 

Is it because we want to make sure we have not forgotten to add migrations?

OskarStark: > Is it because we want to make sure we have not forgotten to add migrations?

Yes

Nyholm: Okey. Fair.

PR #258, opened 2025-03-03
Fix Doctrine ORM deprecations

stof:
This enables the new behaviors of the ORM to avoid getting deprecation warnings.


stof:
We will need to do a dependency update first...

PR #261, opened 2025-04-25
Close issues as 'not_planned'

smnandre:
Following discussion on Slack 

With this PR, both draft PR and stale issues closed by Carson will use the reason 'not_planned'

cf : https://docs.github.com/en/rest/issues/issues?apiVersion=2022-11-28#update-an-issue


stof (approved):

OskarStark (approved):

chalasr (approved):

tucksaun (approved):

mtarld (approved):

alexandre-daubois (approved):

src/Api/Issue/GithubIssueApi.php
```diff
@@ -54,9 +54,21 @@ public function show(Repository $repository, $issueNumber): array
         return $this->issueApi->show($repository->getVendor(), $repository->getName(), $issueNumber);
     }
 
-    public function close(Repository $repository, $issueNumber)
+    public function close(Repository $repository, $issueNumber, ?string $reason = null): void
     {
-        $this->issueApi->update($repository->getVendor(), $repository->getName(), $issueNumber, ['state' => 'closed']);
+        if (null !== $reason && !in_array($reason, [StateReason::COMPLETED, StateReason::NOT_PLANNED], true)) {
```
xabbuh, line 59: What about using an enum to not have to do this check?

smnandre: **In general**, I'm not a big fan of enums used as "closed set of strings" .. as most of the time these sets reveal themselves not to be that closed.

`state_reason` has 3 allowed string values: `completed`, `not_planned` and `reopened`. 

The allowed ones depends on the `state` change:
* `close` -> `open`:  `reopened` or null (defaulting to `reopened`)
* `open` -> `close`:  `completed`, `not_planned` or (defaulting to `completed`)

Meaning an Enum would not be enough to check the value.. or would force us to remove `reopened`, then opening the door to other problems later. 

To be honest I considered two options:
* ❌ implement the whole thing with strong dedicated types/VOs, and make changes in the rest of the codebase to be fully compliant explicit about this semantic / these rules
* ✅ introduce only a couple of constants, more for DX / maintenance reasons (as all our calls to "close" pass "StateReason::NotPlanned" _for now_)

Now if you think it would be better I can make the changes..  (with a `ClosingReason` enum ? -- less 1:1 with the Api.. but would remove this check).

wdyt ?

chalasr: If they're not good at representing closed sets of values, what are they useful for? :) Suggestion makes sense to me here

smnandre: @chalasr any advice regarding the implementation then ? Losing the adherence to the API, or replacing the in_array with constants by an in_array with Enums ?

smnandre: > If they're not good at representing closed sets of values, what are they useful for? 

I meant the opposite sorry..  

Enum are good for closed set of things... not  "closed set of strings" (used quotes with a bit of sarcasm.. should have be more clear)

To illustrate, this is a closed set of things, and the ideal Enum:

```php 
enum CardinalDirection
{
    case North;
    case South;
    case East;
    case West;
}
```

But, from experience, a list of _strings_ that can and will change is not :) 

Most of the times I see enums, i just see what they are in fact used as and used for: strings. 

</opinion>

chalasr: > Most of the times I see enums, i just see what they are in fact used as and used for: strings.

Well that's ok as long as it's a closed set of values i.e. known at compile time. Strings are arbitrary, loosely typed values with no semantic as you pointed out. 

> But, from experience, a list of strings that can and will change is not :)

Every immutable structure may change at some point. One may to add `mid-west` et al to your cardinal positions example if it makes sense in their business. Here we are dealing with the possible states reasons of a system that's part of our business - that's as valid as the days of the week as an enumeration. 

Regarding the implem I let you see what works best - if you want of course :)

smnandre: Of course! Will do tomorrow :)

tucksaun: Looking at the code again, I want to suggest something worth considering IMHO:

All the calls to `GithubIssueApi::close` are passing the same reason (`StateReason::NOT_PLANNED`). So instead of splitting hairs about enums and validation and so on, what if we hard-code the `state_reason` the same way we do for `state`? 🤷

smnandre:
Reverted previous changes (the constants and new signatures) following @tucksaun suggestion in https://github.com/symfony-tools/carsonbot/pull/261#discussion_r2063285544, 

I then added 
* hard-coded "not_planned" reason
* docblock to precise behaviour 
* a test case for GithubIssueApi

wouterj:
While I'm aware that it says "stale" as one of the items in the "Close as not planned" button, are we actually sure users perceive it as such?

In the UI for users, you can only see "Closed as not planned". If I make a bolt estimation, I would say the absolute majority of open source contributors aren't maintainers themselves. And even if they are, they might not be aware of the little gray text underneath the button. I feel like for these users, "not planned" means precisely that: we don't plan on adding this feature.

I think that sends the wrong signal and removes a bit of the welcoming atmosphere that we strive to have in Symfony.

derrabus:
> I feel like for these users, "not planned" means precisely that: we don't plan on adding this feature.

Both closed states might generate that impression. Or even worse: The state we currently use when closing issues could be misinterpreted as "solved". If we feel like we should be more verbose here, we can make Carson explain this in its closing comment.

The "not planned" state is the right sate for issues closed by Carson because it means that this issue has not been fixed and nobody is actively working on it either.

src/Api/Issue/GithubIssueApi.php
chalasr: Indeed :)

xabbuh (approved):

fabpot (approved):

PR #270, opened 2025-04-29
chore(phpstan): set level 8, fix issues + bump PHPStan CI version

smnandre:
Set PHPStan from `level 5` to ~`level 7`~  **`level 8`** 

- add missing return types
- add missing parameters types
- and minor PHPStan issues

 \+ bump PHPStan to `2.1` on the CI

### DIFF
--- a/src/Api/Issue/GithubIssueApi.php
+++ b/src/Api/Issue/GithubIssueApi.php
@@ -3,8 +3,6 @@
 namespace App\Api\Issue;
 
 use App\Model\Repository;
-use App\Service\TaskHandler\CloseDraftHandler;
-use App\Service\TaskHandler\CloseStaleIssuesHandler;
 use Github\Api\Issue;
 use Github\Api\Issue\Comments;
 use Github\Api\Search;
@@ -21,7 +19,7 @@ class GithubIssueApi implements IssueApi
     ) {
     }
 
-    public function open(Repository $repository, string $title, string $body, array $labels)
+    public function open(Repository $repository, string $title, string $body, array $labels): void
     {
         $params = [
             'title' => $title,
@@ -43,7 +41,7 @@ class GithubIssueApi implements IssueApi
         }
     }
 
-    public function lastCommentWasMadeByBot(Repository $repository, $number): bool
+    public function lastCommentWasMadeByBot(Repository $repository, int $number): bool
     {
         $allComments = $this->issueCommentApi->all($repository->getVendor(), $repository->getName(), $number, ['per_page' => 100]);
         $lastComment = $allComments[count($allComments) - 1] ?? [];
@@ -51,18 +49,12 @@ class GithubIssueApi implements IssueApi
         return $this->botUsername === ($lastComment['user']['login'] ?? null);
     }
 
-    public function show(Repository $repository, $issueNumber): array
+    public function show(Repository $repository, int $issueNumber): array
     {
         return $this->issueApi->show($repository->getVendor(), $repository->getName(), $issueNumber);
     }
 
-    /**
-     * Close an issue and mark it as "not_planned".
-     *
-     * @see CloseDraftHandler
-     * @see CloseStaleIssuesHandler
-     */
-    public function close(Repository $repository, $issueNumber): void
+    public function close(Repository $repository, int $issueNumber): void
     {
         $this->issueApi->update(
             $repository->getVendor(),
@@ -78,7 +70,7 @@ class GithubIssueApi implements IssueApi
     /**
      * This will comment on both Issues and Pull Requests.
      */
-    public function commentOnIssue(Repository $repository, $issueNumber, string $commentBody)
+    public function commentOnIssue(Repository $repository, int $issueNumber, string $commentBody): void
     {
         $this->issueCommentApi->create(
             $repository->getVendor(),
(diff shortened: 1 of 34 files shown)


smnandre:
Damn... did not get these errors locally 😶‍🌫️ (will fix tomorrow)

src/Api/Issue/GithubIssueApi.php
```diff
@@ -43,26 +41,23 @@ public function open(Repository $repository, string $title, string $body, array
...
+    /**
+     * @return array<string, mixed>
+     */
+    public function show(Repository $repository, int $issueNumber): array
     {
         return $this->issueApi->show($repository->getVendor(), $repository->getName(), $issueNumber);
     }
 
-    /**
-     * Close an issue and mark it as "not_planned".
```
OskarStark, line 60: Why remove?

src/Api/Label/StaticLabelApi.php
```diff
@@ -13,28 +13,30 @@
...
+        'Locale', 'Lock', 'Mailer', 'Messenger', 'Mime', 'MonologBridge', 'Notifier', 'ObjectMapper',
+        'OptionsResolver', 'PasswordHasher', 'PhpUnitBridge', 'Process', 'PropertyAccess',
+        'PropertyInfo', 'ProxyManagerBridge', 'PsrHttpMessageBridge', 'RemoteEvent', 'Routing',
+        'Scheduler', 'Security', 'SecurityBundle', 'Serializer', 'Stopwatch', 'String',
+        'Templating', 'Translation', 'TwigBridge', 'TwigBundle', 'TypeInfo', 'Uid', 'Validator', 'VarDumper',
+        'VarExporter', 'Webhook', 'WebLink', 'WebProfilerBundle', 'WebServerBundle', 'Workflow',
+        'Yaml',
+    ];
+
     public function getComponentLabelsForRepository(Repository $repository): array
```
OskarStark, line 32: Method needed?

src/Api/Status/GitHubStatusApi.php
```diff
@@ -8,13 +8,19 @@
 
 class GitHubStatusApi implements StatusApi
 {
+    /**
+     * @var array<string, string>
+     */
     private static array $statusToLabel = [
```
OskarStark, line 14: Constant instead?

src/Service/LabelNameExtractor.php
```diff
@@ -13,16 +13,11 @@
  */
 class LabelNameExtractor
 {
-    private static array $labelAliases = [
+    private const array LABEL_ALIASES = [
         'bridge\doctrine' => 'DoctrineBridge',
-        'bridge/doctrine' => 'DoctrineBridge',
```
OskarStark, line 18: Why removing them ?

src/Service/LabelNameExtractor.php
```diff
@@ -57,14 +54,19 @@ public function extractLabels(string $title, Repository $repository): array
...
 
-        $this->logger->debug('Searched for labels in title', ['title' => $title, 'labels' => json_encode($labels)]);
+        $this->logger->debug('Searched for labels in title', ['title' => $title, 'labels' => \json_encode($labels, \JSON_THROW_ON_ERROR)]);
 
         return $labels;
     }
 
-    public function getAliasesForLabel($label)
+    /**
+     * Get all labels for a given label.
```
OskarStark, line 63: The description is confusing

src/Api/Issue/GithubIssueApi.php
smnandre: Hm. Did i mess with my git history while rebasing? That would be a first 😶‍🌫️

src/Subscriber/RewriteUnwantedPhrasesSubscriber.php
```diff
@@ -59,7 +59,10 @@ public function onPullRequest(GitHubEvent $event): void
         ]);
     }
 
-    private function replaceUnwantedPhrases(string $text, &$count)
+    /**
+     * @param int &$count (optional)
```
stof, line 63: should be `@param int|null &$count` with `@param-out int`, to document that it is OK to have a `null` value when calling the method, but that it won't ever be `null` after the call.

stof: and this could even use `int<0, max>|null` as it looks like Psalm wants a more precise type.

smnandre: Thanks!

src/Api/Label/StaticLabelApi.php
smnandre: I see no call indeed, but maybe let's check this in another PR ?

src/Service/LabelNameExtractor.php
smnandre: DX ? Having 5 felt easier to read/maintain, but this was not really in the scope of the PR i must admit. Should i revert ?

fabpot: Let's revert

fabpot: Just read the rest of the code, I think it's fine.

fabpot (approved):

PR #314, opened 2026-09-20
Verify GitHub webhook signatures with SHA-256 instead of SHA-1

fabpot:
Based on #313 for now


OskarStark:
You could use a stacked PR here 😉

fabpot:
> You could use a stacked PR here 😉

Like this?

stof (approved):

